PR #51, opened 2020-04-24
Bug fix: correct the  cell optimisation with constrained ratios

davidbowler:
The original version combined c/a and a/c into a single case,
which gave the wrong behaviour.  Now separated into individual
cases.

### DIFF
--- a/src/move_atoms.module.f90
+++ b/src/move_atoms.module.f90
@@ -4041,6 +4041,8 @@ contains
   !!  CREATION DATE
   !!   30/05/17
   !!  MODIFICATION HISTORY
+  !!   2020/04/24 08:15 dave
+  !!    Bug fix for constrained ratios
   !!  SOURCE
   !!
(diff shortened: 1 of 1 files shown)


davidbowler:
@jackbaker1001 I found that this wasn't working as expected yesterday, and have made a correction - please check.

tsuyoshi38:
As I have not thought of these options (fixed c/a etc.), I don't understand why we need to distinguish fixed c/a and a/c cases.  Although I have not tried the derivation, I can easily imagine that we need some condition between the change of c and change of a.  But, I think we also need some transformation for the direction (a or c, b), from the stress tensor, but do we need to distinguish fixed (c/a) and (a/c) cases ?

davidbowler:
Following the question from @tsuyoshi38 above, I have combined the a/c and c/a (and other) cases.  @tsuyoshi38 @jackbaker1001 can you review please?

jackbaker1001:
I've taken a look and this now looks correct. Fixing c/a and a/c are equivalent, but, I had the search directions wrong!

PR #65, opened 2020-07-13
Hot fix for XC bug with LibXC

### LINKED ISSUE #63: MakeIonFiles segmentation fault for LDA
I have been trying to generate some LDA basis sets and have found that for all elements and basis set sizes MakeIonFiles will SEGFAULT. The error isn't obvious to me, but I have included all files necessary to replicate the issue assuming it is not an issue with my binary or linked libraries. I have only included samples of the issue with H and Si. 

The pseudopotentials and inputs are the same as those found in: https://github.com/OrderN/CONQUEST-release/tree/master/pseudo-and-pao

Without the same Intel compiler, FFTW and LibXC versions I have, do you successfully generate the basis? If so, could you let me know what compilers and libraries you're using? It's possible there is something wrong with my binary or libraries, but I have had success running CONQUEST in the past with these. 

Code versions and compile environment:
- BasisGeneration/Conquest version 1.0.3 (8dc94a8)
- Intel compiler version 2019.3.199
- FFTW version 3.3.8
- LibXC version 4.3.4
- CPU info: Intel Xeon E5-2665

[bad-basis-gen.tar.gz](https://github.com/OrderN/CONQUEST-release/files/4905816/bad-basis-gen.tar.gz)

davidbowler:
Fixes #63

### DIFF
--- a/tools/BasisGeneration/radial_xc_LibXC_module.f90
+++ b/tools/BasisGeneration/radial_xc_LibXC_module.f90
@@ -215,10 +215,11 @@ contains
     real(double), dimension(:), allocatable :: drho_dr, sigma, vrho, vsigma, loc_rho, d2term
 
     flag_energy = .false.
+    ! Necessary for LibXC
+    allocate(exc_array(n_tot))
+    exc_array = zero
     if(PRESENT(exc)) then
        flag_energy = .true.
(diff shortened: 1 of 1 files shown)


smujahed:
This works and compiles

Issue #70, opened 2020-07-18
atomic species wrong in md.frames


negf:
I assume that the 2 column in md.frames for position (force, velocity) is the atomic species and that the atoms are not reordered during the MD run so the atomic species in md.frames should be identical to what is given initially in coord.in, however they are not identical for all atoms. Moreover, comparing md.frames and trajectory.xsf  the coordinates are identical, while the atomic species differs for some atoms.  Also if I look at the structure of md.frames snapshots they clearly make no sense, while the output in trajectory.xsf looks reasonable. 

The version I used here is Git Branch: master; tag, hash: 1.0.3-pre-1-g386a7c. I also attached some sample output. 
[md.tar.gz](https://github.com/OrderN/CONQUEST-release/files/4941111/md.tar.gz)

negf:
I think it is actually not only the output but the forces get accordingly also scaled by the wrong masses and thus the velocities are not calculated correctly. 

And another odd thing is if atoms are fixed in the coordinate file then the forces are set to 0 zero and they don't move during the MD run, but  the velocities are actually not 0 so they have a finite kinetic energy (temperature), does this make sense?

davidbowler:
There are two problems here which have been located by @tsuyoshi38 relating to atom ordering.  (Conquest uses two approaches to atom ordering: partition labelling (related to parallelisation, which changes during a run as atoms move); and global labelling, which is fixed.)  The first problem is in the subroutine `dump_frame`, in the lines:

```fortran
     write(lun,'(a)') "positions"
     call mdl%dump_mdl_atom_arr(lun, mdl%atom_coords)
     write(lun,'(a)') "end positions"
     write(lun,'(a)') "velocities"
     call mdl%dump_mdl_atom_arr(lun, mdl%atom_velocity)
     write(lun,'(a)') "end velocities"
     write(lun,'(a)') "forces"
     call mdl%dump_mdl_atom_arr(lun, mdl%atom_force)
     write(lun,'(a)') "end forces”
```

found here: https://github.com/OrderN/CONQUEST-release/blob/40cf2904315115b8fd38764b7f493b67793fd935/src/md_model_module.f90#L444-L452

For historical reasons (which will be fixed) ``mdl%atom_velocity`` uses partition labelling while the other arrays use global labelling.  This results in the wrong ordering of the velocities.

The second problem is the association of the wrong species array with the ``mdl`` type (it uses ``species`` where it should used ``species_glob``) here:

[...]

PR #74, opened 2020-08-19
Bugfix velocity ordering

tsuyoshi38:
Removed the bug in `md.frames` following the issue #70.
Since arrays related to species and velocities for dumping `md.frames` used partition labeling, we replaced them by those using global labeling (defined in coordinates file). For this, we introduced new arrays `atom_vels`. We are planning to replace `ion_velocity` (partition labeling) by `atom_vels` in other parts. 

In addition to this, 
 a) we now assume   `AtomMove.FixCentreOfMass` is false if some of the atoms are fixed. 
 b) the degree of freedom is reduced by 3, when `AtomMove.FixCentreOfMass` is true.
For a), we will introduce a method to specify the atoms which are used to define the centre of mass.


tsuyoshi38:
I am sorry that I have not worked on this for more than two(!) months.  (partly due to the fracture of my wrist...) 

Two months ago, Dave and I discussed where we should put allocation of atom_vels (newly introduced for atomic velocities with global labeling), and I have moved it from read_atomic_positions to init_md.

(memo for the discussion)
I think we agreed that we want to remove ion_velocity (and use atom_vels) in the future, then move the similar arrays (atom_coord, atom_vels, species_glob) having global labeling from global_module to a new module.  It may be better to introduce a derived type containing these arrays in the new module. 
We also discussed that it may be a good idea that we would use
  1) x(or y or z)_atom_cell  only in electronic structure calculations   and
  2) atom_coord  for the movement of atoms (MD or structure relaxation).
to distinguish the arrays related to partition-labeling or global_labeling atoms.

Paraquat (approved):

davidbowler (approved):

Issue #73, opened 2020-08-18
problem in conquest.bib


tsuyoshi38:
Dr. Watanabe pointed out the problem in writing `conquest.bib`, since `write_bib` in `biblio_module.f90` from all MPI processes.  I also found that the file is sometimes broken.  He also pointed out that CONQUEST sometimes crashes due to this problem.   (It may depend on a system.)  
We can fix it simply by using (if (node == ionode))  in `write_bib`,  I think.

davidbowler:
It can be fixed in `write_bib` or it could be made simpler by changing `compile_biblio` in `references.f90` to make the routine only run for `inode==ionode`.

PR #80, opened 2020-10-07
A small bug in biblio_module

tsuyoshi38:
…io_assign" etc.

In "write_bib", I put "if(inode == ionode)" as the subroutine calls "io_assign" etc.
At present, all processes call io_assign, and try to write information to the file simultaneously, I think.
For debugging, it is also possible to call "write_bib" only from inode. 
(But, I thought it is safer to have if-clause here.)

### DIFF
--- a/src/biblio_module.f90
+++ b/src/biblio_module.f90
@@ -122,6 +122,7 @@ contains
     integer :: lun
     character(len=400)                :: str
 
+   if(inode == ionode) then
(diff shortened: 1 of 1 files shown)


tsuyoshi38:
This is related to the issue #73.

davidbowler (approved):

davidbowler (approved):

PR #82, opened 2020-10-12
Bugfix velocity ordering

tsuyoshi38:
When I moved the allocation part of `atom_vels` from `read_atomic_positions` to `init_md` following Dave's suggestion, I forgot that this array is used (in `update_pos_and_matrices`) even when `md` is not performed.  I now realized that we have to allocate this array  in `update_pos_and_matrices`, in the case of `cg`.
(At present, I assume we should not deallocate `atom_vels` in the subroutine, since this array is used many times during a job. But, if we need to reduce the memory size in other parts, we should deallocate it every time.)


davidbowler (approved):

tsuyoshi38:
> Dave,

Thank you very much for changing the base. I have not considered a possibility that the base was to master, since I just followed the last pull request.  Scary....

PR #86, opened 2021-03-23
Implement correct calculation of rsq

davidbowler:
The bug arose from calculating rsq from undefined variables
rx etc, instead of r(1) etc.  Fixed in two places in non-SC
force routine


davidbowler:
Thanks to @welucas2 for testing this fix which removes the segfault seen previously with Cray compiler

lionelalexandre (approved):

tsuyoshi38 (approved):

PR #91, opened 2021-05-12
Introduced a module for libXC version 5

tsuyoshi38:
Introduced the module for libXC version 5, following Watanabe-san's code.
SCAN, rSCAN, r2SCAN can be used only with version 5.


davidbowler:
I don't understand why there is a change in these lines:

https://github.com/OrderN/CONQUEST-release/blob/01de36b8d0aa3f921d95c690a1200f297853f849/src/XC_LibXC_v5_module.f90#L663-L672

In the other modules, the d_e_xc/d n calculation is after the select statement, and there is no use of the flag_is_GGA, for instance (and line 650 above):

https://github.com/OrderN/CONQUEST-release/blob/01de36b8d0aa3f921d95c690a1200f297853f849/src/XC_LibXC_v4_module.f90#L715-L721

Do you know why this was changed?

tsuyoshi38:
As you probably imagine, it is partly because we are planning to introduce metaGGA case in this "select" part.
One reason for the change is the lengthy part after calling xc_f90_gga_exc_vxc in the case (XC_FAMILY_GGA) is common, necessary also in the case of metaGGA. 
Another reason is I thought it is better to finish "select" sentences earlier, in general. 
(But, this is my personal preference.)

Do we need to introduce this change later ?  (when we introduce metaGGA)

davidbowler:
The select clause is now shortened with `flag_is_GGA` used to discriminate (added this to `get_libxc_dpotential` as well).

davidbowler (approved):

PR #95, opened 2021-07-30
Fix DFT energy

davidbowler:
Implement new subroutines to find the Hartree, XC and local pseudopotential/neutral atom energy from the output density so that the formal definition of DFT energy is calculated.  Also remove the check_DFT parameter.  Some tidying associated with unnecessary calls to `get_energy`.

### DIFF
--- a/src/H_matrix_module.f90
+++ b/src/H_matrix_module.f90
@@ -1739,6 +1739,8 @@ contains
   !!  CREATION DATE
   !!   2021/07/22
   !!  MODIFICATION HISTORY
+  !!   2021/07/26 10:54 dave
+  !!    Added local PS for nonNA runs, tidied output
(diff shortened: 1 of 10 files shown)


davidbowler:
Note that the extensive changes to the XC modules are simply to implement subroutines that calculate the XC energy (without the potential).

tsuyoshi38 (approved):

tsuyoshi38:
For 32H2O system,
1st DFT energy with Neutral Atom 
 -563.906805699888309  vs -563.906805699882170 (new)
without Neutral Atom
 -563.906769206129866   vs -563.906769205682622 (new)

Sorry, I have noticed that the result in the former was obtained by a rather old version. (Still the agreement is satisfactory)

PR #98, opened 2021-09-24
Merge stabilised quasi-Newton method (SQNM) for structural relaxation

davidbowler:
SQNM behaves much better than existing structural relaxation
approaches, particularly when forces are small.  It should be 
considered the optimal structural relaxation method.

NB documentation is missing from this pull request at the moment
- I will add it before merging.


davidbowler:
Documentation (relatively minimal) added.

PR #102, opened 2021-10-20
Create write_extxyz() for output a .xyz file with extended format

JianboLin:
The extended format type xyz file can attach information not only atomic positions, but also lattice vectors, energy, forces, etc.
The benefit of this format is that it can be recognized by Python Library ASE and very convenient to do machine learning with such data.
With ASE, it is easy to do further structure analysis or converting to other format of geometry, also preparation of Machine learning.
Add config_type (crystal_Si, liquid_Si, amorphous_Si, mono_vacancy_Si, etc.) to comment line from IO.Title with which people can easily separate or select the geometries in the same xyz file.

### DIFF
--- a/src/control.f90
+++ b/src/control.f90
@@ -1420,7 +1420,7 @@ subroutine update_pos_and_box(baro, nequil, flag_movable)
   subroutine write_md_data(iter, thermo, baro, mdl, nequil)
 
     use GenComms,      only: inode, ionode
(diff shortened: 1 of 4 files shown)


davidbowler:
I added two comments to explain the changes (the main one is for the move of the titles variable) and added a save attribute to titles to make it persist.

PR #105, opened 2021-11-01
Unify the units in write_extxyz()

JianboLin:
Since the distance unit in xyz file is Angstrom, we expect the units of force and energy are eV/Angstrom and eV.
I feel sorry that the present units of force and energy are still Ha/Bohr and Ha. It is because the convert value of for_conv is 1.0 in default, then it will not change force unit when we use default units in Conquest_input.
Now, some local parameters of for_conv_loc, en_conv_loc, dist_conv_loc are used to
make all the units in this xyz file are with Angstrom (position) , eV (energy) and eV/Angstrom (force).

### DIFF
--- a/src/io_module.f90
+++ b/src/io_module.f90
@@ -2719,6 +2719,8 @@ second:   do
   !!   Writes atomic positions, including atomic forces, lattice vectors, 
   !!  energy, system signature, etc. to an extended format .xyz file   
   !!  which can be directly recognized by Python Library named ASE.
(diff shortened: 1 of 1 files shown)


tsuyoshi38 (approved):

davidbowler (approved):
This makes sense: since the output expects eV and Angstroms, the local conversion variables are a good solution.

JianboLin:
Thank you for your confirmation.

Issue #110, opened 2022-05-27
Indexing errors during L-BFGS relaxation on multiple processes


davidbowler:
See discussion #109 for full details; in brief, running some L-BFGS relaxations on more than one process causes indexing errors.  The problem seems to come from the variable `g0` being uninitialised on processes that are not `ionode`.

davidbowler:
Further insight: in`control.f90`,  `g0` was defined on `ionode` but not on other processes on the first iteration, so at l. 2116 we had the potential for the search direction to be redefined on all processes apart from `ionode` leading to problems with indexing as different processes moved identical atoms to different positions:

https://github.com/OrderN/CONQUEST-release/blob/e0f11dc46ffc73c58b315f355625824b599873cf/src/control.f90#L2113-L2120

Defining `g0` at l. 2017 will fix this.

https://github.com/OrderN/CONQUEST-release/blob/e0f11dc46ffc73c58b315f355625824b599873cf/src/control.f90#L2017

davidbowler:
Fixed by #111

PR #111, opened 2022-05-27
Correct error in lbfgs defining g0

davidbowler:
The variable g0 was only defined on ionode on the first iteration
of lbfgs, and not on other processes.  This led to different search
directions and indexing errors (principally i_h2d).  The variable
is now defined on all processes.

This fixes issue #110 (discussed in #109 )

Bug reported by @lionelalexandre and root cause found by @negf


lionelalexandre (approved):
Approved !

PR #128, opened 2022-12-02
F rework output ase

lionelalexandre:


lionelalexandre:
By setting "IO.WriteOutToASEFile T" in Conquest input (default is F), Conquest will generate an output named "Conquest_out_ase" containing minimum data (species/eigenstates/DFT energy/forces/stress) for ASE. This output is 
** independent of Conquest output printing option 
** generated dynamically thus avoiding accumulation of data
The code is in a minimal/hacked form it will need more clever writing in future

PR #132, opened 2022-12-04
F rework output debug lionel

lionelalexandre:
**Correction for some printing format
**Spin polarisation setup with NeUp NeDn
**ASE kpoint writing (fractional coordinates are mandatory)
Tested of non-polarized and polarised band of structures/DOS of Fe using ASE.
Please review these modifications and, if you have time, test if spin-polarisation is working fine (Tsuyoshi has reported some issues.

Have a nice week !


davidbowler (approved):

davidbowler:
Only one small change: removed the '==zero' test and replaced it with 'abs()<RD_ERR' to avoid floating point rounding errors

PR #166, opened 2023-02-24
F md analyser

Lorentyfle:
**Edition log:**

Addition of the dynpro program created by L.Truflandier.
To set up dynpro, use the command on the terminal "make dynpro" inside the cq_dynpro1.0 folder.

The python code was adapted to send the files in the good format inside different folders to avoid the possibility to edit the original input files.

For now, only the RDF, the VACF, the PSD were added.
No verification was bring for the MSD, however the diffusion coefficient was added.

The RDFs and PSD are the only ones that can be ploted with the python code.
For the MSD, the diffusion coefficient, the RDF and PSD it is now possible to print in function of specific species or atoms.
It could be either with the ID or the species.
It is important to note that the IDs for the diffusion coefficient/MSD and for the RDF/PSD are different!
The first one see the IDs in the md.frames files, and the second one look up at the ID along sub.struct_perm.lk.

The coordination number given for each pics need to be reworked. The PSD can only go up to 2^100 frames.
Visual bugs are corrected for the plot of the md.stats file if the option "compare" isn't selected.

The speed for the calculation of the RDF and the PSD were improved.
Comments were added on the python code.


lionelalexandre:
Only the md-analyser python files can be pushed ; the dynpro code can't be added there.

Issue #173, opened 2023-04-06
Post-processing issue for band density: integration value


JianboLin:
When we integrate the value from the cube file of band density and multiply with grid volume, it is expected to result 1.
But, I find it is not always 1. 
This problem was detected in SiGe system 1000 atoms. Since this SiGe system is large, I prepared with small systems of 
1) 2 Fe atoms BCC, hamman DZP PBE
2) 8 Si atoms diamond, hamman DZP PBE
3) 8 Si atoms diamond, hamman DZP LDA

All the total charge densities (chden.cube) give correct results of electron number.
In the case of Fe, integrations of band density give correct values near to 1. However, in two Si cases, integrations of band density give strange values far from 1.
Please find the files of .txt including these results. There are also input files for these three small systems.
I hope you have your own code to do the integration. If not, please use the cube_tool of python code in analysis_tool, with cmd,
python cube_tools_v2.0_original.py -i  filename.cube

I can not upload cube files, only 25MB allowed.
[PostProcess_chden_bandden_check.tar.gz](https://github.com/OrderN/CONQUEST-release/files/11169931/PostProcess_chden_bandden_check.tar.gz)

davidbowler:
I think that this will be addressed in branch bugfix-post-process-charge

JianboLin:
Thank you very much. I tested using the same examples with this branch of bugfix-post-process-charge. The results of total charge density keep the previous value. The results of band density show no significant improvement with present version.

davidbowler:
I just ran a test on the `Si8_diamond_DZP_hamman_LDA` case, and the integral is 1.000 in every case.  Did you rebuild the post-processing tool? (something like `cd tools/PostProcessing; make`).  Once you have rebuilt it, when it runs you should get a line like this in the output:

`Git Branch: bugfix-post-process-charge; tag, hash: v1.0.5-pre-289-g8fd5a861`

Please check the Git hash (it should be 8fd5a861).

JianboLin:
Thank you for your comment! I rebuilt PostProcess and recalculated the examples from begining. 
Now all the three cases give 1.000 for the integral value.

davidbowler:
That's excellent! Could you approve the pull request, please? (Or maybe @tsuyoshi38 or @ayakon could approve it)

davidbowler:
Fixed by #181

Issue #172, opened 2023-03-14
Post-processing issue for charge density with small cells


davidbowler:
If the radius of a PAO is larger than the cell size, then during post-processing to produce charge density or band-resolved densities the code will crash.  (Fix is straightforward in `pao_to_grid`)

davidbowler:
Fixed by #181

PR #181, opened 2023-05-19
Bugfix post process charge

davidbowler:
This should fix the problems with bands not integrating to one (also fixed a problem with wrapping grid points).

These are problems reported in issues #173 and #172


tsuyoshi38 (approved):
I don't understand some of the details (related to i_mult) now, but it looks correct with ol_ang_coeff_subs.

PR #182, opened 2023-05-23
Implement contiguous access to acc_block in get_matrix_elements

davidbowler:
Change GEMM call in get_matrix_elements (transpose) so that access to acc_block is contiguous within innermost loop, and replace inner loop with array segment.


lionelalexandre (approved):

tkoskela (approved):
Profiling on Vtune showed 6x better performance in the inner loop. I trust you've checked the results are still correct. Of course you can wait until #184 is merged and the CI will tell you :smile:

davidbowler:
Results appear to be fine, so I'm merging.

PR #183, opened 2023-05-23
Remove if statements from get_matrix_elements

davidbowler:
The inner loops of get_matrix_elements contain many checks on indexing which slow the code down.  These have been removed.


tkoskela:
Should we close this? I expect it will cause conflicts with #182

davidbowler:
Either close it or rework it once #182 has been merged

davidbowler:
I have fast-forwarded this onto develop so it's now a very small set of changes which I think can be merged without any problems.  It would be interesting to see how much difference it makes.

src/calc_matrix_elements_module.f90
```diff
@@ -207,7 +207,6 @@ subroutine get_matrix_elements_new(myid,rem_bucket,matM,gridone,gridtwo)
           allocate(acc_block(n_dim_two*n_dim_one),STAT=stat)
           if(stat/=0) call cq_abort('Error allocating memory in get_matrix_elements: ',n_dim_one,n_dim_two)
           acc_block = zero
-          call reg_alloc_mem(area_integn,n_dim_two*n_dim_one,type_dbl)
```
tkoskela, line 210: Is removing this related to the if statments?

tkoskela (commented):
Looks clean to me. I had assumed most of these if statements were already removed in #182 but I'm happy to remove them with this PR. Just one question about the memory allocations below

src/calc_matrix_elements_module.f90
davidbowler: Not strictly, no - I was maybe a little over-zealous in tidying! I can put this back if we want and remove it in a separate PR...

davidbowler: (I have removed a call to a subroutine which registers the amount of memory allocated and the equivalent subroutine which notes that it's been freed, which is really not needed here; it doesn't actually allocate anything)

tkoskela (approved):

PR #189, opened 2023-05-24
Tk/ci skeleton

### LINKED ISSUE #186: Add CI workflow skeleton
In order to run the tests added in #177 automatically, we need to set up a CI workflow for the develop branch. I'm going to start by adding a basic skeleton that will compile the code. Once #184 is merged we can add a simple test that just checks the code didn't crash. Then once we start writing tests with a testing framework those can be added in as additional steps.

tkoskela:
Closes #186 

This PR adds a simple CI workflow to the `develop` branch. At the moment it just checks out the code and builds it. We can add more tests once we have them.


davidbowler:
Ubuntu includes libXC I think (at least based on a quick search) - I would think that it would be useful to include this in the tests as well.  I'm assuming it would be easy to add.

tkoskela:
The name is a bit ambigous, do you mean [this package](https://launchpad.net/ubuntu/+source/libxc)? I can add it

davidbowler:
Yes - that's the package - thank you for making it more specific!

davidbowler (approved):

tkoskela:
There is an error from `git symbolic-ref HEAD` during the build
```
sed "s/BBB/"`git symbolic-ref HEAD | cut -b 12-`"/" verstr.f90 | sed "s/RRR/"`git describe --tags`"/" >> datestamp.f90
10
fatal: ref HEAD is not a symbolic ref
```
This is because the action runs on a merge commit,  not a branch, as explained in [this issue](https://github.com/peter-evans/create-pull-request/issues/163). I assume parsing the branch name fails, but since this only runs in the GHA runner, I don't think we should mind too much.

Issue #162, opened 2023-02-14
Poor SCF convergence at small residuals with Kerker pre-conditioning


davidbowler:
In many cases during SCF with Kerker pre-conditioning, if the SCF tolerance is tighter than 1e-7 or so, we can find very slow convergence (typically the convergence to 2e-7 is fast and then it simply oscillates).  This seems to be caused by the zeroing of the q=0 component of the Fourier transform of the residual in `hartree.f90`.  See attached example, a Pt(111) surface, for an example.

[SCFResidualOscillation.zip](https://github.com/OrderN/CONQUEST-release/files/10732894/SCFResidualOscillation.zip)

negf:
I have seen similar issues with systems which are not periodic in 3D. I had
the feeling that the problem is that the Kerker pre-conditioner assumes PBC
in 3D. Essentially due the finite cut-off of the basis functions the charge
in the vacuum region (away from the surface) should be identical to 0.
However due to the FFT (for the Kerker pre-conditioner) one gets a finite
charge density there.

davidbowler:
Yes, I think that's one aspect that's a problem.  Fundamentally, the q=0 component should be zero (it represents the integral over all space of the residual) and if it's not then setting it to zero causes problems near SCF with tight tolerances.

If you'd like to test the fix, the branch is `bugfix-poor-scf-convergence` and I've just pushed a commit.

negf:
Sure, I will give it a try. Thanks.

davidbowler:
This is now fixed by #194

PR #194, opened 2023-06-05
Bugfix poor scf convergence

davidbowler:
Fixes issues identified in #162


lionelalexandre (approved):
Did not get time to check for this. I guess you have tested it.

tsuyoshi38 (approved):
This change should be okay. In principle, output and input charge density should have the same electron number and its change should be 0 in the case of spin =1. 
(though there seems some effects)
(I think I have removed this constraint for spin=2 case, since the term for q=0 should vary during the SCF cycle in this case. But, now I cannot find when and where I have done this..)

PR #195, opened 2023-06-09
Thread loops over blocks

### LINKED ISSUE #178: Add threading over blocks in calc_matrix_elements

tkoskela:
PR to close #178 

This PR adds threading over the biggest loops over blocks in `calc_matrix_elements_module` and `PAO_grid_transform_module`.

In `calc_matrix_elements_module` threading is done in subroutines `get_matrix_elements_new` and `act_on_vectors_new`. Both subroutines have a deep loop nest where the loop over blocks is the outermost loop. In `get_matrix_elements_new` the innermost loop accumulates data in `send_array`, therefore a reduction is done at the end of the parallel region. In `get_matrix_elements_new`, `gridfunctions%griddata` is updated with a `gemm` call in the innermost loop. The updates are done in separate indices by each loop iteration, therefore it can be operated on by all threads in parallel.

In `PAO_grid_transform_module`, the `single_PAO_to_grid` subroutine is rewritten so that threading over blocks can be done. The innermost loop updates `gridfunctions%griddata`. However the index was calculated sequentially inside the loop, which made it unsafe for threading. The rewritten subroutine first precomputes the indices and stores them in an array, then does the loop over blocks which can now safely be threaded. The `single_PAO_to_grad` subroutine has not been refactored, it will be removed in a separate PR and the functionality merged with `single_PAO_to_grid` to address #244


tkoskela:
I just realized the values in the output are only written out at 8 decimals. So my race condition is not producing any observable error. Very strange :thinking:

davidbowler:
I have tested the attached file (216 atoms which scales happily enough to 9 MPI processes) on combinations from 1 to 8 MPI processes with 1 to 4 OpenMP threads and found the same energy to 10 decimal places for all cases.  I'm not sure if this helps!

[LiCltest.tgz](https://github.com/OrderN/CONQUEST-release/files/11745920/LiCltest.tgz)

tkoskela:
I think this PR is now ready to merge to development. We can address #244 in a separate PR to avoid cluttering this one further. Here is a performance comparison to development. I ran these on young using 8 MPI ranks and the attached input files
[input_data.zip](https://github.com/OrderN/CONQUEST-release/files/12426913/input_data.zip). 

|   | develop | This PR  |
|---|---|---|
| Without `-fopenmp`                      | 95.331s  | 63.482s |
| With `-fopenmp`, 1 OMP thread   | 95.597s  | 62.324s |
| With `-fopenmp`, 4 OMP threads |   | 36.491 s |

Even the serial performance is significantly improved! I've started work on the code for #244 and it looks like we can gain a bit more speedup in threaded performance.

davidbowler:
Running the LiCl test (input files above) on a local cluster (AMD, Intel compiler 2020_u4, MKL, OpenMPI 4.1.3) I get a segfault when using two threads and 9 MPI processes (the fault seems to come from a reduction).  Another segfault occurs using GCC 11.2.0 on the same system.  On a Mac MBP M2 with GCC 12.2 and OpenMPI 4.1.5 the same test runs without problems.

davidbowler:
I find the same problem with your test above

tkoskela:
Interesting, is it possible to get access to the cluster?

On young I'm using 
- `compilers/intel/2022.2`
- `mpi/intel/2021.6.0/intel`

gcc and openmpi are available there as well, I can see if I can build with them

davidbowler:
The problem comes from `calc_matrix_elements_new` (found by trial and error - commenting out different areas) but I don't know more yet.

davidbowler:
We are now confident that the problem here is coming from the OpenMP stack (the errors I found were caused by a stack overflow...).  Setting the environment variable OMP_STACKSIZE fixes this issue (in particular, I found that a value somewhere around 50M was helpful though this might require a little testing).

I will add to the documentation before approving the PR.

docs/installing.rst
```diff
@@ -66,3 +66,12 @@ Once these are set, you should make the executable using ``make``.
 
 The ion file generation code is compiled using the same options
 required for the main code.
+
+Multi-threading
+~~~~~~~~~~~~~~~
+
+CONQUEST can use OpenMP for multi-threading; some multi-threading is available throughout the code, while there are specific matrix multiplication routines which can use multi-threading for the linear scaling solver.  The number of threads is set via the environment variable ``OMP_NUM_THREADS``.
+
+Compiler flags to enable OpenMP are dependent on the vendor, but should be specified via ``COMPFLAGS`` in the ``system.make`` file.  If compiling with OpenMP then you should also change the variable ``OMP_DUMMY`` in the same file to be blank to enable the number of threads to be included in the output.
```
tkoskela, line 75: Small comment here: The OpenMP flag also needs to go into ``LINKFLAGS``

davidbowler (approved):

src/PAO_grid_transform_module.f90
```diff
@@ -107,136 +107,145 @@ subroutine single_PAO_to_grid(pao_fns)
...
-                endif
-
-                xatom=DCS_parts%xcover(icover)
-                yatom=DCS_parts%ycover(icover)
-                zatom=DCS_parts%zcover(icover)
-                the_species=species_glob(ig_atom)
+    ! Note: Using OpenMP in this loop requires some redesign because there is a loop
+    !       carrier dependency in next_offset_position.
+    blocks_loop: do iblock = 1, domain%groups_on_node ! primary set of blocks
+       part_in_block: if(naba_atoms_of_blocks(atomf)%no_of_part(iblock) > 0) then ! if there are naba atoms
```
tkoskela, line 174: This may not be necessary, if `naba_atoms_of_blocks(atomf)%no_of_part(iblock) == 0`, the loop below will not execute?

PR #199, opened 2023-06-23
Implement Resta approach to polarisation

davidbowler:
Implemented Resta approach to modern theory of polarisation, added new section to manual on electronic structure analysis (including polarisation), added test for simple BTO cell.


davidbowler:
@tkoskela Could you check that my implementation of a test which only applies to this polarisation case is correct please?

testsuite/test_check_output.py
```diff
@@ -43,6 +47,10 @@ def test_check_outputs(test_path, key):
     # Template for skipping tests. If a parameter combination is added to
     # xfail_test, pytest will expect the test to fail.
     xfail_test = (key == "Not-a-real-key")
+    if (xfail_test):
+        pytest.xfail("invalid parameter combination: "+test_path+", "+key)
+    # Only check polarisation for test003 for now
+    xfail_test = (key == "Total polarisation" and not(test_path[7]=="3"))
```
tkoskela, line 53:
```suggestion
    xfail_test = (key == "Total polarisation" and not("test_003" in test_path))
```

tkoskela (changes requested):
The implementation of the test looks ok to me. I've made a suggestion to make the xfail check on the test path a bit clearer but otherwise it should be fine.

The issue is that the [Total stress does not match the reference value](https://github.com/OrderN/CONQUEST-release/actions/runs/5354629135/jobs/9711906337#step:11:98) on GitHub actions, nor on my laptop, to four decimal precision.

On my laptop it's very close to passing
```
Max absolute difference: 0.00016935
Max relative difference: 3.90025266e-06
 x: array([-43.4698, -43.4698, -43.4204])
 y: array([-43.4699, -43.4699, -43.4203])
```

whereas on Github actions the errors are fairly large

```
Max absolute difference: 0.00281102
Max relative difference: 6.47353749e-05
 x: array([-43.4698, -43.4698, -43.4204])
 y: array([-43.4723, -43.4723, -43.4232])
```
I'm starting to wonder if we need to take a bit closer look at what is going on in the total stress calculation. This magnitude of errors is a bit concerning.

tkoskela:
We can also remove the `not-a-real-key` example from the testsuire, since we now have a real case of keys not found in some of the tests.

tkoskela (approved):
Approved with my suggsted changes. The stress test even passes now

PR #201, opened 2023-06-23
Refactor the testsuite

### LINKED ISSUE #200: Refactor output checks in testsuite
We want to include rather than exclude keys from the output

tkoskela:
Closes #200

Each Fortran test now has its own parametrized python test that specifies the values it checks from the output. I would have liked to also make `results` a fixture, but I couldn't figure out how to pass the path into a fixture.

### DIFF
--- a/testsuite/README.md
+++ b/testsuite/README.md
@@ -22,11 +22,6 @@ These steps can be run automatically using the script `run_conquest_test.sh` in 
 
 To add new tests
 
-  1. Add input files and a sample output file (run with `IO.Iprint 0` and named `Conquest_out.ref`) in a new subdirectory under [testsuite](./). The naming convention is test directory names start with `test_` followed by a running index with three digits, e.g. `003`.
-  2. Add an entry to the `"test_path"` parameter in [test_check_output.py](./test_check_output.py). The test driver will check for all fields in the output listed in the `"key"` parameter. They are read from the [Conquest_out](test_001_bulk_Si_1proc_Diag/Conquest_out.ref) file by the `read_conquest_out()` function.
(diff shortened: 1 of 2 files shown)


davidbowler:
Can you update the README.md to reflect how we now create new tests, and in particular how we can specify a different precision to the default?

davidbowler (approved):

tkoskela:
> Can you update the README.md to reflect how we now create new tests, and in particular how we can specify a different precision to the default?

Thanks, good point! I've done it, I hope this has made it a bit simpler to add tests as well.

In the future, you can select "request changes" in the review if you would like to request changes :smiley:

davidbowler:
Can you add an example for how to set a precision that is not the default (or add it to the README.md)?

tkoskela:
To set a custom precision I would maybe just modify the precision matching the key in the test function. Something like:

```
    @pytest.mark.parametrize("key", ['Harris-Foulkes energy',
                                     'Max force',
                                     'Force residual',
                                     'Total polarisation'])
    def test_003(self, key, default_precision, testsuite_directory):

        path = os.path.join(testsuite_directory, "test_003_bulk_BTO_polarisation")
        res = results(path, key)
        if key is 'Total polarisation':
          precision = custom_precision
        else
          precision = default precision
        np.testing.assert_allclose(res[0], res[1], rtol = precision, verbose = True)
```
Doing it through the fixtures seems too complicated

davidbowler:
Could we also parametrize the precision in the same way as the key?

tkoskela:
Yeah, I just realised we could. Then it can't be a fixture but I don't think that will matter.

tkoskela:
You already merged this PR, do you want me to open another one?

PR #206, opened 2023-06-30
Implemented user choice for number of atoms in pDOS

davidbowler:
User can now specify a number of atoms and equivalent atom numbers (index in input file).  Defaults to all atoms.


ayakon (approved):

ayakon:
I tried the code and it worked appropriately!

PR #207, opened 2023-07-03
Add selecting custom precision by key

tkoskela:
From discussion of #201. This PR allows you to pass a key to the precision function and return a different precision.


davidbowler (approved):

tkoskela:
Let's merge this tomorrow once we've had a look at it

PR #210, opened 2023-07-05
Change OMP parallel loop to remove errors

davidbowler:
The no_of_ib_ia variable is determined within the loop and isn't thread-safe (as far as I can see) so I've moved the OpenMP loop further in which in my tests is actually slower with multi-threading though I can't see why!

This is only a demonstration, not code to be merged or used in production


tkoskela:
I think we can close this, I cherry-picked your commits from this branch into `tk-thread-calc-matrix` and the code is now refactored in #245

Issue #208, opened 2023-07-04
fermi_offset for STM is not used in PostProcessing


ayakon:
"fermi_offset" can be read as 'STM.FermiOffset' in Conquest_input
but is never used in the code.
It should be used when we set Emin and Emax in stm_module.f90.

davidbowler:
Closed by #216

PR #216, opened 2023-07-18
PP: Add Fermi_offset

ayakon:
"fermi_offset" can be read as 'STM.FermiOffset' in Conquest_input
but was never used in the code.
The code has been modified to add fermi_offset to Emin and Emax in stm_module.f90.

This branch is related to issue #208 .

### DIFF
--- a/tools/PostProcessing/stm_module.f90
+++ b/tools/PostProcessing/stm_module.f90
@@ -349,6 +349,8 @@ contains
     allocate(current(nptsx,nptsy,nptsz))
     allocate(psi(nptsx,nptsy,nptsz))
 
+    if (nspin==1 .and. fermi_offset.ne.zero) &
+       write(*,fmt='(2X,"Fermi-offset is ",f7.3," eV")') fermi_offset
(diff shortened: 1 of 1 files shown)


davidbowler (approved):
This is fine, but is there a reason that we write out Fermi offset only when `nspin==1`? Since we apply the offset whatever the spin setting, it might be better to write it out in all cases.

tools/PostProcessing/stm_module.f90
```diff
@@ -348,16 +348,21 @@ subroutine tersoff_hamann
     call set_prefac_real(9)
     allocate(current(nptsx,nptsy,nptsz))
     allocate(psi(nptsx,nptsy,nptsz))
+
+    if (nspin==1 .and. fermi_offset.ne.zero) &
```
davidbowler, line 352: Maybe remove `nspin==1` here

ayakon: Thank you for your comment.
I removed nspin==1
and also changed the digit of the output from f7.3 to f10.3.

Issue #209, opened 2023-07-04
Unit of Efermi in Conquest_out of PostProcessing


ayakon:
In Conquest_out of PostProcessing, Fermi level is shown in Hartree unit
(ex., "Fermi level:     -0.17719 Ha").
while the STM energy region is given in eV unit
(ex., "Including bands between  -4.822 and  -4.722 eV", where -4.822 eV = -0.17719 Ha.),
so it is not easy to compare with Fermi level.

It might be good that we show Fermi level both in Hartree and eV units,
somehow like "Fermi level:     -0.17719 Ha  =  -4.822 eV".

davidbowler:
Closed by #217

PR #217, opened 2023-07-18
PP: Added output of Efermi in eV unit

ayakon:
In Conquest_out of PostProcessing, Efermi is written only in Hartree unit,
while the energy range of STM is written in eV.

So for easy comparison with the energy range, Efermi in eV unit has been added in Conquest_out.

This is related to issue #209 .

### DIFF
--- a/tools/PostProcessing/read_module.f90
+++ b/tools/PostProcessing/read_module.f90
@@ -458,10 +458,10 @@ contains
     efermi = zero
     if(nspin==1) then
        read(17,fmt='(a6,f18.10)') str,efermi(1)
-       write(*,fmt='(4x,"Fermi level: ",f12.5," Ha   (=",f7.3," eV)")') efermi(1), efermi(1)*HaToeV
+       write(*,fmt='(4x,"Fermi level: ",f12.5," Ha   (=",f10.3," eV)")') efermi(1), efermi(1)*HaToeV
     else
        read(17,fmt='(a6,2f18.10)') str,efermi(1), efermi(2)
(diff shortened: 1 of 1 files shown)


tools/PostProcessing/read_module.f90
```diff
@@ -458,10 +458,10 @@ subroutine read_eigenvalues
     efermi = zero
     if(nspin==1) then
        read(17,fmt='(a6,f18.10)') str,efermi(1)
-       write(*,fmt='(4x,"Fermi level: ",f12.5," Ha")') efermi(1)
+       write(*,fmt='(4x,"Fermi level: ",f12.5," Ha   (=",f7.3," eV)")') efermi(1), efermi(1)*HaToeV
```
davidbowler, line 461: I wonder about the formatting here: we might need more digits before the decimal point when using eV than using Ha.  Should we change it to `f10.3` maybe?

tools/PostProcessing/read_module.f90
```diff
@@ -458,10 +458,10 @@ subroutine read_eigenvalues
...
     else
        read(17,fmt='(a6,2f18.10)') str,efermi(1), efermi(2)
-       write(*,fmt='(4x,"Fermi levels: ",2f12.5," Ha")') efermi
+       write(*,fmt='(4x,"Fermi levels: ",2f12.5," Ha   (=",2f7.3" eV)")') efermi, efermi*HaToeV
```
davidbowler, line 464: Same formatting question here: maybe `f10.3` might be safer.

davidbowler (approved):

tools/PostProcessing/read_module.f90
ayakon: Thank you for your comment.
I modified them to f10.3.

PR #218, opened 2023-07-19
Fixed problems with non-SCF forces and structural optimisation

### LINKED ISSUE #118: Problems with non-SCF forces and structural optimisation
Structural optimisation can fail when using the non-SCF approach.  There appear to be at least three issues:

1. `AtomMove.ReuseSFcoeff` is set to `T` by default, which forces `AtomMove.ReuseDM` to `T`, so that the charge density is built from the output density matrix *after* the new atomic density has been created.  This just needs a careful consideration of default values.
2. The forces calculated with the non-SCF approach appear to be incompatible with the energy (in the attached phenol example the error in z on atom 3 is around 10% for example).
3. The stress seems incorrect (likely the Hartree stress)

I strongly suggest not using non-SCF calculations until these can be fixed.  Issue found by @lionelalexandre

davidbowler:
There was a missing minus sign in the PCC part of the non-SCF forces, and a missing non-SCF exception when rebuilding the charge density from the DM after atomic movement.

With this change, non-SCF forces are correct, and non-SCF stress is correct for the non-neutral-atom case (another fix is needed for that!).

Fixes #118

### DIFF
--- a/src/force_module.f90
+++ b/src/force_module.f90
@@ -3412,6 +3412,8 @@ contains
   !!    Added atomic stress contributions
   !!   2021/03/23 15:52 dave
   !!    Bugfix: correct calculation of rsq implemented, variables tidied
+  !!   2023/07/19 10:40 dave
+  !!    Added minus sign to non-SCF PCC force for consistency with non-SCF part
   !!  SOURCE
   !!
   subroutine get_nonSC_correction_force(HF_force, density_out, inode, &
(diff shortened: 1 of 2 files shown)


lionelalexandre (approved):
that's great !

davidbowler:
That's very helpful to know, Tsuyoshi: thank you for checking.

PR #222, opened 2023-07-20
Implementation of Hamiltonian padding

tsuyoshi38:
This is for the discussion (draft pull request): 
For implementing "padding H and S matrices", flag_padH and matrix_size_padH have been introduced.
Any suggestion for the naming of variables ?


src/DiagModule.f90
```diff
@@ -574,7 +574,7 @@ subroutine FindEvals(electrons)
        ! node, note that the repeating of the same eigenvalues in each
        ! proc_group is taken care of by the additional factor
        ! 1 / N_procs_in_pg
-       call gsum(w(:,:,spin), matrix_size, nkp)
+       call gsum(w(:,:,spin), matrix_size_padH, nkp)
     end do ! spin
```
davidbowler, line 578: Do we really need the sum over `matrix_size_padH` here? We don't really want the padded eigenvalues in the end, so it might be better just to sum over `matrix_size`

davidbowler (commented):
I will look in more detail at this as soon as I can

src/DiagModule.f90
tsuyoshi38: I just thought that since this is the size of the 1st rank (changing 1st), it is safer to use the same size defined for this array. I will consider it tomorrow (check actual gsum for 2-rank array), but in any case, the speed should be almost the same, I guess.

tsuyoshi38: By the way, I think I have already closed this pull request. Should I restore it?
(Anyway, thank you very much for your review and comment. > Dave)

davidbowler: You haven't closed it - it's still a draft, and I haven't approved it!

tsuyoshi38: I see.
 I misunderstood that this is a pull request for merging `f-proj_PHM_DiagModule` to `f-proj_PadHamiltonianMatrix`. 
Yes. I would like to keep the discussion on "Padding H matrix" here.

tsuyoshi38: >Do we really need the sum over matrix_size_padH here? 
>We don't really want the padded eigenvalues in the end, so it might be better just to sum over matrix_size

I first thought that this may be correct. But, I have checked `gsum` again and I think my first reply is correct.
In `double_two_gsumv` in `generic_comms.f90`, the first dummy argument (`variable`) is defined in the following way, 

  subroutine double_two_gsumv(variable, len1, len2)
    use datatypes
    implicit none
    ! Passed variables
    integer :: len1, len2
    real(double), dimension(len1,len2) :: variable

As far as I learned, since only the first address of the variable is passed in the fortran, the size of the first dimension needs to be consistent between the calling part and within this subroutine (double_two_gsumv). 

If we would introduce an additional passed variables in this subroutine, we can make the summation only for actual matrix. But, I don't think it necessary.

tsuyoshi38: If your suggestion is about the size of `w`, we may be able to reduce its size.  I will check it.
(To be honest, I want to use `ww` instead of `w`.  It is difficult to search a variable, if its name is just expressed by one character. Similarly, I prefer to use "ii" instead of "i", even if it would be used only within do-loops.)

davidbowler: I think that you're right; the size is needed for the correct use of gsum.  I agree that w is a very poorly named variable (sorry!) but I think that a more descriptive name (like `eigenvalues`) would be better than `ww`.

tsuyoshi38: I have changed the name of `w` as `evals`.  
Following this change, `local_w` was renamed as `local_evals`.
(Sorry, `local_eigenvalues` seemed long to me.)

tsuyoshi38: I forgot to report...
I have reduced the size of `w` (`evals`). 
Actually, its size must be `matrix_size` to be consistent with the subroutines (like `write_output_ase`).
Although this array was used in the first call of `pzhegvx`, we can use `local_w` (`local_evals`) instead for this call.

JianboLin:
> This is for the discussion (draft pull request):
> For implementing "padding H and S matrices", flag_padH and matrix_size_padH have been introduced.
> Any suggestion for the naming of variables ?

Sorry, Miyazaki-san, I clicked the button of 'ready for review' by mistake.

tsuyoshi38:
???  Lin-san, I don't understand what you have done.
Have you issued a pull request for `f-proj_PadHamiltonianMatrix` to develop ?

JianboLin:
> ??? Lin-san, I don't understand what you have done.
> Have you issued a pull request for `f-proj_PadHamiltonianMatrix` to develop ?

Sorry, I think I only changed the status of this pull request to **[Ready for review]**. 
Can I click **[Convert to draft]** under 'Reviewers' at the right side of this page to return to the previous status?

davidbowler (approved):

PR #227, opened 2023-07-24
Pull request to proj_PadHamiltonianMatrix (should work with a given block size)

tsuyoshi38:
I have not checked many examples, but I would like to create a pull request to the branch (proj_PadHamiltonianMatrix). 
For my test systems, I have confirmed that this version correctly runs with a given block size provided in Conquest_input.
I think this version should be already very useful, since it should improve the efficiency of CONQUEST dramatically in some cases. (especially when the size of Hamiltonian or Overlap matrix is a prime number)
At present, I assume that users will provide the block size by theirselves. 

1. I have not considered the case for the small size of matrix.   (when block size ~ matrix size)
2. I have not checked whether the code works for multiple k-points, with `Diag.KProcGroups` more than 1.
3. I have not followed how (# of proc rows) and (# of proc cols) are determined automatically in the present version of the code yet. 
4. It is probably necessary to have a default block size, which needs to be automatically set from the number of MPI processes and matrix size.

All of these problems will be discussed in another issue.

### DIFF
--- a/src/DiagModule.f90
+++ b/src/DiagModule.f90
@@ -2136,12 +2136,8 @@ contains
                    ! localEig(Distrib%firstrow(recv_proc+1)+j-1,coff:coff+block_size_c-1) = RecvBuffer(j,k:k+block_size_c-1)
 
                    if(coff+block_size_c-1 > matrix_size) then 
-                    !TMP localEig(coff:matrix_size,Distrib%firstrow(recv_proc+1)+j-1) = RecvBuffer(j,k:k+block_size_c-1)
-         
-                    do l=1, block_size_c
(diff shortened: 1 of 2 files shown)


tsuyoshi38:
I have checked the following case, 
2. I have not checked whether the code works for multiple k-points, with Diag.KProcGroups more than 1.
and have found a bug in this case.

I will fix it soon.

tsuyoshi38:
I have restored the branch, revised the code in this branch and confirmed the revised version runs correctly. 
But, I don't know how I can issue a pull request for the merge from this branch (`f-proj_PHM_DiagModule`) to `f-proj_PadHamiltonianMatrix`.

PR #228, opened 2023-07-28
A bug in the case where  Diag.KProcGroups > 1 has been removed.

tsuyoshi38:
A bug in the case where  Diag.KProcGroups > 1 has been removed. 
Size of w in gsum is also changed following the comment by Dave.


tsuyoshi38:
> Size of w in gsum is also changed following the comment by Dave.
This was not correct without changing the size of 'w'.

PR #237, opened 2023-08-03
Bug(?) in the if-statement before calling dump_pos_and_matrices

tsuyoshi38:
It may not be a bug, but ..
after I updated the MacOS, I encountered a problem in the following part with the option (-O3).

if (n_dumpL>0 .and. mod (n_iter, n_dumpL) == 0) then 
 call dump_pos_and_matrices(index=unit_DM_save)
endif

The problem seems to come from the 2nd one when n_dumpL=0. I thought it is not used since we also checked n_dumpL. But, it is probably safer to change it to

if (n_dumpL>0) then
 if(mod (n_iter, n_dumpL) == 0) call dump_pos_and_matrices(index=unit_DM_save) 
endif

### DIFF
--- a/src/DMMinModule.f90
+++ b/src/DMMinModule.f90
@@ -1235,8 +1235,8 @@ contains
        end if
 
        ! dump the L matrix if required
-       if (n_dumpL>0 .and. mod (n_iter, n_dumpL) == 0) then
-           call dump_pos_and_matrices(index=unit_DM_save)
+       if (n_dumpL>0) then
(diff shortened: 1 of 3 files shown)


tsuyoshi38:
I am sorry, but I have found that even after I made these changes the code does not run with -O3. But, CQ runs correctly with -g. On our supercomputer, CQ runs correctly in both cases, with and without these changes.

davidbowler:
Even if this doesn't fix the problem you reported with -O3, it's still a very good change to have made! This is much safer and better coding practice.

davidbowler (approved):

Issue #244, opened 2023-08-16
Reduce code duplication in PAO_grid_transform_module


tkoskela:
The subroutines 

https://github.com/OrderN/CONQUEST-release/blob/1378f0359b798362b84177bc4288e97ceff17824/src/PAO_grid_transform_module.f90#L98

and

https://github.com/OrderN/CONQUEST-release/blob/1378f0359b798362b84177bc4288e97ceff17824/src/PAO_grid_transform_module.f90#L270

Duplicate almost all of the code with only minor differences. It would be good for code stability to move the common code into a utility function, or perhaps merge the subroutines into one. Worth keeping in mind if they need refactoring for #198

davidbowler:
This is a good point.  We could easily merge the two and add a flag to select PAO value or gradient to choose whether to call `evaluate_pao` or `pao_elem_derivative_2` (though this latter name should be rationalised for clarity!)

tkoskela:
The simplest way is to put an `if` clause in the innermost loop. I've got a draft version of this in [tk-reduce-pao-duplication](https://github.com/OrderN/CONQUEST-release/tree/tk-reduce-pao-duplication). I'm a bit concerned about the performance implication of an if statement in the innermost loop. I would like to spend a bit of time to explore different options.

tkoskela:
Procedure pointers in Fortran 2003 could be useful here. According to a comment in [this thread](https://stackoverflow.com/questions/8612466/how-to-alias-a-function-name-in-fortran) the performance gains can be dubious, but worth checking in my opinion. If there is no real performance gain, it becomes more of a question of which coding style we prefer. I don't have a strong opinion on which style is better, another if-layer in the loop nest or an obscure function interface.

davidbowler:
I wonder if it's not easier to just use an abstract interface and pass the required subroutine as an argument, as described in [this thread](https://stackoverflow.com/questions/32809769/how-to-pass-subroutine-names-as-arguments-in-fortran)

tkoskela:
> I wonder if it's not easier to just use an abstract interface and pass the required subroutine as an argument, as described in [this thread](https://stackoverflow.com/questions/32809769/how-to-pass-subroutine-names-as-arguments-in-fortran)

Yes, that does look easier! Thanks, I wasn't aware you could pass subroutines without pointers

PR #245, opened 2023-08-22
Improve threading in PAO_grid_transform_module

tkoskela:
The threading implementation in PAO_grid_transform_module currently in `tk-thread-calc-matrix` has a lot of OpenMP overheads, as the vtune analysis below shows. This PR is an attempt to thread the loop over blocks as we originally envisioned, by moving the calculation of offsets to `griddata` into a separate loop. It significantly improves the performance of the threaded region starting around 10s:

image

to

image

I'd recommend using this as a starting point for implementing #244

### DIFF
--- a/src/PAO_grid_transform_module.f90
+++ b/src/PAO_grid_transform_module.f90
@@ -107,7 +107,7 @@ contains
     use species_module, ONLY: species, npao_species
     !  At present, these arrays are dummy arguments.
     use block_module, ONLY : nx_in_block,ny_in_block,nz_in_block, n_pts_in_block
-    use group_module, ONLY : blocks, parts
+    use group_module, ONLY : blocks
     use primary_module, ONLY: domain
     use cover_module, ONLY: DCS_parts
     use set_blipgrid_module, ONLY : naba_atoms_of_blocks
@@ -121,7 +121,7 @@ contains
     !local
     integer :: iblock,ia,ipart,ip,l1,acz,m1 ! Loop index variables
     integer :: nblock, npart, natom ! Array dimensions
-    integer :: jpart,ind_part,ii,icover,ig_atom,iatom ! Temporary indices for calculating species
+    integer :: naba_part_label,ind_part,icover,iatom ! indices for calculating species. TODO:Need better names
     integer :: position,next_offset_position ! Temporary indices to gridfunctions%griddata
     integer :: my_species ! Temporary variables to reduce indirect accesses
     integer :: npoint ! outputs of check_block
@@ -130,11 +130,11 @@ contains
     real(double) :: x,y,z ! Temporary variables to reduce indirect accesses
     real(double) :: rcut ! Input to check_block
     real(double) :: val ! output, written into gridfunctions%griddata
-    real(double) xblock,yblock,zblock ! inputs to check_block
+    real(double) :: xblock,yblock,zblock ! inputs to check_block
+    real(double) :: xatom,yatom,zatom ! inputs to check_block
     integer, allocatable, dimension(:) :: ip_store ! outputs of check_block
-    integer, allocatable, dimension(:,:,:) :: offset_position, species_store ! precomputed variables
+    integer, allocatable, dimension(:,:,:) :: offset_position ! precomputed offsets
     real(double), allocatable, dimension(:) :: x_store, y_store, z_store, r_store ! outputs of check_block
-    real(double), allocatable, dimension(:,:,:) :: xatom,yatom,zatom ! inputs to check_block, precomputed
 
     nblock = domain%groups_on_node
     npart = maxval(naba_atoms_of_blocks(atomf)%no_of_part)
@@ -142,15 +142,11 @@ contains
 
     call start_timer(tmr_std_basis)
     call start_timer(tmr_std_allocation)
-    allocate(xatom(natom, npart, nblock))
-    allocate(yatom(natom, npart, nblock))
-    allocate(zatom(natom, npart, nblock))
     allocate(ip_store(n_pts_in_block ))
     allocate(x_store( n_pts_in_block ))
     allocate(y_store( n_pts_in_block ))
     allocate(z_store( n_pts_in_block ))
     allocate(r_store( n_pts_in_block ))
-    allocate(species_store(    natom, npart, nblock))
     allocate(offset_position(natom, npart, nblock))
     call stop_timer(tmr_std_allocation)
     ! --  Start of subroutine  ---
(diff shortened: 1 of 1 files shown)


src/PAO_grid_transform_module.f90
```diff
@@ -115,127 +115,147 @@ subroutine single_PAO_to_grid(pao_fns)
...
                 iatom=iatom+1
                 ii = naba_atoms_of_blocks(atomf)%list_atom(iatom,iblock)
                 icover= DCS_parts%icover_ibeg(jpart)+ii-1
                 ig_atom= id_glob(parts%icell_beg(ind_part)+ii-1)
 
-                xatom=DCS_parts%xcover(icover)
-                yatom=DCS_parts%ycover(icover)
-                zatom=DCS_parts%zcover(icover)
-                the_species=species_glob(ig_atom)
+                xatom(      ia, ipart, iblock) = DCS_parts%xcover(icover)
```
davidbowler, line 193: I don't think that we need to store xatom, yatom and zatom or species_store here - better to calculate them in the loop below

src/PAO_grid_transform_module.f90
```diff
@@ -115,127 +115,147 @@ subroutine single_PAO_to_grid(pao_fns)
...
-                   !$omp             shared(n_pts_in_block, pao, the_species, offset_position, &
-                   !$omp                    npoint, no_of_ib_ia, ip_store, r_store, x_store, y_store, z_store) &
-                   !$omp             private(r_from_i, ip, position, ipoint, &
-                   !$omp                     x, y, z, l1, acz, m1, count1, val)
-                   do ip=1,npoint
-                      ipoint=ip_store(ip)
-                      position= offset_position + ipoint
-
-                      r_from_i = r_store(ip)
+                call check_block (xblock(iblock), yblock(iblock), zblock(iblock), &
```
davidbowler, line 220: Again, I don't think that we need xblock etc to have dimension iblock here

src/PAO_grid_transform_module.f90
```diff
@@ -115,127 +115,147 @@ subroutine single_PAO_to_grid(pao_fns)
...
-                            do m1=-l1,l1
-                               call evaluate_pao(the_species,l1,acz,m1,x,y,z,val)
-                               !gridfunctions(pao_fns)%griddata(position+(count1-1)*n_pts_in_block) = val
-                               temp_block_storage(ipoint + (count1-1)*n_pts_in_block) = val
+                      count1 = 0
+                      l_loop: do l1 = 0,pao(my_species)%greatest_angmom
+                         z_loop: do acz = 1,pao(my_species)%angmom(l1)%n_zeta_in_angmom
+                            m_loop: do m1=-l1,l1
+                               call evaluate_pao(my_species,l1,acz,m1,x,y,z,val)
+                               gridfunctions(pao_fns)%griddata(position + count1 * n_pts_in_block) = val
```
davidbowler, line 242: We could remove count1 here and just increment position by n_pts_in_block at each loop step I think

src/PAO_grid_transform_module.f90
```diff
@@ -115,127 +115,147 @@ subroutine single_PAO_to_grid(pao_fns)
...
+    nblock = domain%groups_on_node
+    npart = maxval(naba_atoms_of_blocks(atomf)%no_of_part)
+    natom = maxval(naba_atoms_of_blocks(atomf)%no_atom_on_part)
 
     call start_timer(tmr_std_basis)
     call start_timer(tmr_std_allocation)
-    allocate(ip_store(n_pts_in_block),x_store(n_pts_in_block),y_store(n_pts_in_block),z_store(n_pts_in_block), &
-         r_store(n_pts_in_block))
+    allocate(xblock(nblock), yblock(nblock), zblock(nblock))
+    allocate(xatom(natom, npart, nblock))
```
davidbowler, line 146: xblock etc, xatom etc, species_store should not be arrays I think

src/PAO_grid_transform_module.f90
tkoskela: Yes, I think you are right! They are not needed for the calculation of the offsets

src/PAO_grid_transform_module.f90
tkoskela: You are definitely right about `xblock`, `yblock`, `zblock`. We only need the species to calculate the offsets. I have moved those to the openmp loop. With `xatom`, `yatom`, `zatom` and `species_store`, it depends on whether we want to redo the calculation of `icover` and `ig_atom` in the parallel loop. Maybe that would make sense, it won't be that expensive and will save us memory allocations.

src/PAO_grid_transform_module.f90
tkoskela: Well spotted! That makes it a lot clearer

src/PAO_grid_transform_module.f90
tkoskela: I've changed the code so that only the offsets are stored in an array. It does look much better, thanks for the feedback! I moved the index calculations that need to be repeated to get `my_species` in both loops to a separate routine to avoid duplication.

src/PAO_grid_transform_module.f90
```diff
@@ -107,135 +107,145 @@ subroutine single_PAO_to_grid(pao_fns)
...
+    !       carrier dependency in next_offset_position.
+    blocks_loop: do iblock = 1, domain%groups_on_node ! primary set of blocks
+       part_in_block: if(naba_atoms_of_blocks(atomf)%no_of_part(iblock) > 0) then ! if there are naba atoms
+          iatom = 0
+          parts_loop: do ipart=1,naba_atoms_of_blocks(atomf)%no_of_part(iblock)
+             naba_part_label = naba_atoms_of_blocks(atomf)%list_part(ipart,iblock)
+             ind_part = DCS_parts%lab_cell(naba_part_label)
+             atoms_loop: do ia=1,naba_atoms_of_blocks(atomf)%no_atom_on_part(ipart,iblock)
+
+                call get_species(iblock, naba_part_label, ind_part, iatom, icover, my_species)
```
davidbowler, line 181: What do we gain by creating the get_species subroutine? I'm slightly nervous about the (hidden) increment of iatom and icover at a lower level, but happy to approve if there are good reasons

tkoskela: The subroutine is called in both loops so we gain not having to write the same code twice. 

I suppose we can increment `iatom` outside of the subroutine if that feels safer, although since we are explicitly passing it into a subroutine as `intent(inout)`, modifying it feels like fair game to me. `icover` is `intent(out)` so the subroutine just assigns it a value, similar to `my_species`.

src/PAO_grid_transform_module.f90
```diff
@@ -107,135 +107,145 @@ subroutine single_PAO_to_grid(pao_fns)
...
-                ig_atom= id_glob(parts%icell_beg(ind_part)+ii-1)
-
-                xatom=DCS_parts%xcover(icover)
-                yatom=DCS_parts%ycover(icover)
-                zatom=DCS_parts%zcover(icover)
-                the_species=species_glob(ig_atom)
+    ! Note: Using OpenMP in this loop requires some redesign because there is a loop
+    !       carrier dependency in next_offset_position.
+    blocks_loop: do iblock = 1, domain%groups_on_node ! primary set of blocks
+       part_in_block: if(naba_atoms_of_blocks(atomf)%no_of_part(iblock) > 0) then ! if there are naba atoms
```
tkoskela, line 174: Do we need this if statement? If `naba_atoms_of_blocks(atomf)%no_of_part(iblock) == 0`, the following loop will not happen

PR #251, opened 2023-08-29
Combine single_PAO_to_grid and single_PAO_to_grad

### LINKED ISSUE #244: Reduce code duplication in PAO_grid_transform_module

tkoskela:
Closes #244 

I've merged the two identical subroutines into `PAO_or_gradPAO_to_grid`. The subroutine called in the inner loop is passed in as a procedure argument through an interface. As a side effect I had to always pass `direction` into both subroutines to make them conform to the same interface, in `evaluate_pao` it's just a dummy argument that does nothing. I've updated all calls to `PAO_or_gradPAO_to_grid`, `evaluate_pao` and `pao_elem_derivative_2` and added a bit of explanation in comments.

Very open to a more descriptive names for the subroutine and the interface

Gets rid of OpenMP overheads in `single_PAO_to_grad` we were seeing earlier

## Previously:
image

## This PR:
image

### DIFF
--- a/src/PAO_grid_transform_module.f90
+++ b/src/PAO_grid_transform_module.f90
@@ -114,13 +114,11 @@ contains
     use primary_module, ONLY: domain
     use cover_module, ONLY: DCS_parts
     use set_blipgrid_module, ONLY : naba_atoms_of_blocks
-    !use angular_coeff_routines, ONLY : evaluate_pao, pao_elem_derivative_2
     use functions_on_grid, ONLY: gridfunctions, fn_on_grid
     use pao_format
 
     implicit none
     integer, intent(in) :: pao_fns
-    !procedure(evaluate_interface), intent(in) :: evaluate
     integer, intent(in) :: direction
 
     !local
@@ -143,7 +141,7 @@ contains
 
     interface
        ! Interface to return a value val given arguments
-       ! direction,species,l1,acz,m1,x,y,z. Implemented by
+       ! direction,species,l,acz,m,x,y,z. Implemented by
        ! evaluate_pao() and pao_elem_derivative_2().
        subroutine evaluate(direction,species,l,acz,m,x,y,z,val)
          use datatypes, only: double
--- a/src/PAO_grid_transform_module.f90
+++ b/src/PAO_grid_transform_module.f90
@@ -54,10 +54,10 @@ module PAO_grid_transform_module
 
 contains
 
-!!****f* PAO_grid_transform_module/single_PAO_to_any *
+!!****f* PAO_grid_transform_module/PAO_or_gradPAO_to_grid *
 !!
 !!  NAME
-!!   single_PAO_to_any
+!!   PAO_or_gradPAO_to_grid
 !!  USAGE
 !!
 !!  PURPOSE
@@ -95,10 +95,10 @@ contains
 !!    Removed unused npao1 and atom_species
 !!   2023/08/23 11:20 tkoskela
 !!    Added OMP threading, merged single_PAO_to_grad and single_PAO_to_grid into
-!!    single subroutine single_PAO_to_any (for lack of better name)
+!!    single subroutine PAO_or_gradPAO_to_grid
 !!  SOURCE
 !!
-  subroutine single_PAO_to_any(pao_fns, evaluate, direction)
+  subroutine PAO_or_gradPAO_to_grid(pao_fns, evaluate, direction)
 
     use datatypes, only: double
     use primary_module, ONLY: bundle
@@ -258,7 +258,7 @@ contains
     call stop_timer(tmr_std_allocation)
     call stop_timer(tmr_std_basis)
     return
-  end subroutine single_PAO_to_any
+  end subroutine PAO_or_gradPAO_to_grid
   !!***
 
 ! -----------------------------------------------------------
--- a/src/PAO_grid_transform_module.f90
+++ b/src/PAO_grid_transform_module.f90
@@ -121,7 +121,7 @@ contains
     implicit none
     integer, intent(in) :: pao_fns
     !procedure(evaluate_interface), intent(in) :: evaluate
-    integer, intent(in), optional :: direction
+    integer, intent(in) :: direction
 
     !local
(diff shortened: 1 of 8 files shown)


tkoskela:
I don't see any noticeable performance impact of passing the subroutine as an argument instead of selecting it with an `if` statement inside the loop nest. I like not having an `if` statement at the bottom of the loop nest from a code readability point of view.

src/PAO_grid_transform_module.f90
```diff
@@ -111,12 +114,14 @@ subroutine single_PAO_to_grid(pao_fns)
...
-    use angular_coeff_routines, ONLY : evaluate_pao
+    !use angular_coeff_routines, ONLY : evaluate_pao, pao_elem_derivative_2
     use functions_on_grid, ONLY: gridfunctions, fn_on_grid
     use pao_format
 
     implicit none
-    integer,intent(in) :: pao_fns
+    integer, intent(in) :: pao_fns
+    !procedure(evaluate_interface), intent(in) :: evaluate
+    integer, intent(in), optional :: direction
```
davidbowler, line 124: I don't think that direction should be optional: it's always passed to evaluate.

src/PAO_grid_transform_module.f90
```diff
@@ -54,10 +54,10 @@ module PAO_grid_transform_module
 
 contains
 
-!!****f* PAO_grid_transform_module/single_PAO_to_grid *
+!!****f* PAO_grid_transform_module/single_PAO_to_any *
```
davidbowler, line 57: The name here isn't really quite right, but I can't think of an alternative!  The previous names (`single_PAO_to_grid` and `single_PAO_to_grad`) but also misleading (the second should have been `single_PAO_grad_to_grid`).

I wonder if `single_PAO_to_grid` would be better though it might confusing historically

davidbowler (changes requested):

src/PAO_grid_transform_module.f90
tkoskela: I don't understand the algorithm well enough to give the subroutine a descriptive name, so I'm really waiting for your input here :sweat_smile:  If you think `single_PAO_to_grid` is good, let me know.

davidbowler: Possibly `single_PAO_or_grad_PAO_to_grid` which is at least descriptive! I don't think that we need the `single` part, so maybe `PAO_or_grad_PAO_to_grid`.  @tsuyoshi38 and @ayakon do you have any thoughts?

src/PAO_grid_transform_module.f90
```diff
@@ -111,12 +114,14 @@ subroutine single_PAO_to_grid(pao_fns)
     use primary_module, ONLY: domain
     use cover_module, ONLY: DCS_parts
     use set_blipgrid_module, ONLY : naba_atoms_of_blocks
-    use angular_coeff_routines, ONLY : evaluate_pao
+    !use angular_coeff_routines, ONLY : evaluate_pao, pao_elem_derivative_2
```
ilectra, line 117: Remove if not needed anymore

src/PAO_grid_transform_module.f90
```diff
@@ -136,6 +141,19 @@ subroutine single_PAO_to_grid(pao_fns)
     integer, allocatable, dimension(:,:,:) :: offset_position ! precomputed offsets
     real(double), allocatable, dimension(:) :: x_store, y_store, z_store, r_store ! outputs of check_block
 
+    interface
+       ! Interface to return a value val given arguments
+       ! direction,species,l1,acz,m1,x,y,z. Implemented by
```
ilectra, line 146:
```suggestion
       ! direction,species,l,acz,m,x,y,z. Implemented by
```

src/PAO_grid_transform_module.f90
```diff
@@ -111,12 +114,14 @@ subroutine single_PAO_to_grid(pao_fns)
...
     use set_blipgrid_module, ONLY : naba_atoms_of_blocks
-    use angular_coeff_routines, ONLY : evaluate_pao
+    !use angular_coeff_routines, ONLY : evaluate_pao, pao_elem_derivative_2
     use functions_on_grid, ONLY: gridfunctions, fn_on_grid
     use pao_format
 
     implicit none
-    integer,intent(in) :: pao_fns
+    integer, intent(in) :: pao_fns
+    !procedure(evaluate_interface), intent(in) :: evaluate
```
ilectra, line 123: Remove if not needed anymore

ilectra (approved):
Only minor aesthetic comments

tkoskela:
> Only minor aesthetic comments

Thanks! All done now

src/PAO_grid_transform_module.f90
tsuyoshi38: `PAO_or_grad_PAO_to_grid` sounds reasonable to me.  (I would like to remove `_` between grad and PAO, like `PAO_or_gradPAO_to_grid`

davidbowler (approved):

src/PAO_grid_transform_module.f90
tkoskela: Technically this should be "any function that satisfies the `evaluate` interface to grid" but I suppose we can name it after the current implementations. Would `transform_to_grid` be too generic?

davidbowler: I think that would risk confusion as this routine is specifically for PAOs and their derivatives (there are other functions transformed to the grid).  I've taken what @tsuyoshi38 and I agreed (`PAO_or_gradPAO_to_grid`) and pushed it.

PR #253, opened 2023-08-30
Implementation of variable temperature

AugustinLu:


src/control.f90
```diff
@@ -824,6 +826,23 @@ subroutine md_run (fixed_potential, vary_mu, total_energy)
 
     do iter = i_first, i_last ! Main MD loop
        mdl%step = iter
+       
+       ! At a given time step, if the difference between T_ext and the final temperature is larger than 
+       ! the temperature step (with 5% margin), then update T_ext 
+       ! Temperature evolves linearly from temp_ion to temp_ion_end over MDn_steps steps
+       ! Stop when target temperature has been reached
+       if  (abs(mdl%T_ext - temp_ion_end) > 0.95*abs((temp_ion_end-temp_ion)/dble(MDn_steps))) then
```
davidbowler, line 834: I wonder if we should make this fraction of 0.95 a parameter that can be set: presumably it acts as some kind of thermal inertia/

src/control.f90
```diff
@@ -824,6 +826,23 @@ subroutine md_run (fixed_potential, vary_mu, total_energy)
...
+         mdl%T_ext = temp_ion+(dble(iter-1)/dble(MDn_steps))*(temp_ion_end-temp_ion)
```
davidbowler, line 835: Can we define `(temp_ion_end - temp_ion)/real(MDn_steps,double)` as a variable such as `temp_change_step`? Please use `real( ,double)` rather than `dble()` for portability.

src/control.f90
```diff
@@ -824,6 +826,23 @@ subroutine md_run (fixed_potential, vary_mu, total_energy)
...
+         write(io_lun,fmt='(6x, "Thermostat temperature at step ", i5, ": ", f9.1, " K")') iter, mdl%T_ext 
+         ! Update target ke for SVR
+         thermo%ke_target = half*md_ndof_ions*fac_Kelvin2Hartree*mdl%T_ext
+         write(io_lun,fmt='(6x, "kee target is now" , f8.3)') thermo%ke_target
```
davidbowler, line 839: This needs to be only on `ionode` (or `if(myid==0)` and only when `iprint` is high enough

src/control.f90
```diff
@@ -824,6 +826,23 @@ subroutine md_run (fixed_potential, vary_mu, total_energy)
...
+         
+       else if (abs(temp_ion_end-temp_ion) > 0.0001) then
+         write(io_lun,fmt='(6x, "Target temperature (", f9.1," K) has been reached. Stopping..")') mdl%T_ext
```
davidbowler, line 842: This needs to be only on `ionode` (or `if(myid==0)` and only when `iprint` is high enough. Please also make the 0.0001 parameter adjustable

src/initial_read_module.f90
```diff
@@ -2690,6 +2691,9 @@ subroutine write_info(titles, mu, vary_mu, HNL_fac, NODES)
        call chrcap(ensemblestr,3)
        write(io_lun, fmt='(4x,a15,a3," MD run for ",i5," steps ")') job_str, ensemblestr, MDn_steps
        write(io_lun, fmt='(6x,"Initial ion temperature: ",f9.3,"K")') temp_ion
+       if (temp_ion_end .ne. temp_ion) then
+         write(io_lun, fmt='(6x,"Final ion temperature: ",f9.3,"K")') temp_ion_end
```
davidbowler, line 2695: This needs to be only on `ionode` (or `if(myid==0)` and only when `iprint` is high enough

src/control.f90
AugustinLu: Actually, it is a (possibly clumsy) way to check at a given time step whether we have reached the target final temperature or not. 
With float numbers, setting the condition at exactly the temperature step, or to zero, it happened that the condition was not properly fulfilled and the cooling/heating continued beyond the target temperature.

src/control.f90
AugustinLu: Should it be based on variable `iprint_gen` or `iprint_MD` ?

$\rightarrow$ `if(inode==ionode.AND.iprint_MD>2)` ?

davidbowler: Since it's related to MD, I think that `iprint_MD` is better

src/control.f90
JianboLin: Sorry for cutting in. But, I also would like to make this option of changing temperature available soon.
For the updating of target temperature(`mdl%T_ext`),
I have tested some melt-quench calculations with the following way.  It will shift the target temperature from `mdl%T_ext` to `temp_ion_end` with `temp_change_step` K for each md step. 
```
       temp_diff = temp_ion_end - mdl%T_ext
       if (abs(temp_diff) > temp_change_step) then
         mdl%T_ext = mdl%T_ext + temp_change_step * abs(temp_diff)/temp_diff
       else
         mdl%T_ext = temp_ion_end
       end if
```

AugustinLu: Than you for your comment. 
I would like to point out or ask the following:

- `temp_change_step` may be negative (ex: cooling) so the condition `(abs(temp_diff) > temp_change_step)`. Thus we should also use the absolute value of the temperature step in the condition. 
- If we restart the calculation, will the temperature be appropriately set by your code snippet? It had some issues with it before, so this is why I set the thermostat temperature directly as a linear function of the current step number `iter`.

src/control.f90
```diff
@@ -824,6 +826,23 @@ subroutine md_run (fixed_potential, vary_mu, total_energy)
...
     do iter = i_first, i_last ! Main MD loop
        mdl%step = iter
+       
+       ! At a given time step, if the difference between T_ext and the final temperature is larger than 
+       ! the temperature step (with 5% margin), then update T_ext 
+       ! Temperature evolves linearly from temp_ion to temp_ion_end over MDn_steps steps
+       ! Stop when target temperature has been reached
+       if  (abs(mdl%T_ext - temp_ion_end) > 0.95*abs((temp_ion_end-temp_ion)/dble(MDn_steps))) then
+         mdl%T_ext = temp_ion+(dble(iter-1)/dble(MDn_steps))*(temp_ion_end-temp_ion)
+         write(io_lun,fmt='(6x, "Thermostat temperature at step ", i5, ": ", f9.1, " K")') iter, mdl%T_ext
```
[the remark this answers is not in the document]
AugustinLu: The following is proposed:
```
         ! Report the current thermostat temperature
         if (inode==ionode .and. iprint_MD > 0) then
            write(io_lun,fmt='(6x, "Thermostat temperature at step ", i5, ": ", f9.1, " K")') iter, mdl%T_ext
         end if
         if (inode == ionode .and. iprint_MD > 1 ) then
            write(io_lun,fmt='(6x, "kee target is now" , f8.3)') thermo%ke_target
         end if
```

The current thermostat temperature is reported on  `ionode` if `iprint` > 0 and the target kinetic energy is reported if `iprint` > 1

src/control.f90
AugustinLu: I realize that setting this small number was unnecessary. Simply checking that `temp_ion/=temp_ion_end` should be sufficient to trigger this exit command.

```
       if [...]
         [...]
       else if (temp_ion_end /= temp_ion) then ! Ensure that temp_ion_end was different from temp_ion
         if (inode == ionode) then
           write(io_lun,fmt='(6x, "Target temperature (", f9.1," K) has been reached. Stopping..")') mdl%T_ext
         end if
		 exit
       end if
```

davidbowler: It's never safe to use `/=` to compare two real numbers.  We have a parameter `RD_ERR` in `numbers` module that would work here:

```
else if(abs(temp_ion_end - temp_ion)>RD_ERR) then...
```

would be much better.

AugustinLu: I see! Then I will replace the condition containing hard-coded 0.0001 with RD_ERR.

src/control.f90
JianboLin: > * `temp_change_step` may be negative (ex: cooling) so the condition `(abs(temp_diff) > temp_change_step)` would be always true. Thus we should also use the absolute value of the temperature step in the condition.

I think a positive value of `temp_change_step` is good, because `temp_ion_end` gives a direction from current temperature for cooling or heating.

> * If we restart the calculation, will the temperature be appropriately set by your code snippet? It had some issues with it before, so this is why I set the thermostat temperature directly as a linear function of the current step number `iter`.

In my test, when restart, I set `AtomMove.ReadVelocity ` to True, and update `temp_ion`, so CQ will read velocity and make thermostat from this `temp_ion`. 
In develop branch, `iter` number will restart from beginning. 

> * There is a difference in behaviour  between the current implementation and the proposed one for simulations with restarts. What should happen when the final temperature is reached?
>   
>   1. Exit as soon as the temperature reaches `temp_ion_end`? The simulation then run exactly the number of steps required by the input file, regardless of whether a restart has occurred.
>   2. Continue and maintain the temperature at `temp_ion_end`? The number of steps actually run will depend on when restarts have occurred.

For a cooling or heating process, after `mdl%T_ext`  reaches `temp_ion_end`,  it may be good to exit.
which means after completing the rest part in this loop.

AugustinLu:
I accidentally changed the section on Berendsen thermostat and barostat in the documentation file, so I reverted these changes to only keep those related to the variable temperature functionality in 06826b6d29b22e0a143726b6c269f6d52f9af865

tsuyoshi38:
Please resolve the conflict shown above. (you just need to introduce a line for multiply_kernel, I think.)

src/md_control_module.f90
```diff
@@ -2503,11 +2511,62 @@ subroutine write_md_checkpoint(th, baro)
         end if
       end if
       call io_close(lun)
+
+      if (flag_variable_temperature) then
+        ! Write instantaneous temperature in md.temperature
+        open(unit=lun,file=md_temperature_file,status='replace')
```
tsuyoshi38, line 2517: Here, before opening unit=lun, you should do 
   call io_assing(lun)

src/md_misc_module.f90
```diff
@@ -69,15 +69,15 @@ module md_misc
   !!    (introduced atom_vels in July, 2020.)
   !!   2022/10/04 17:27 dave
   !!    Reworking to set initial KE of ions correctly
```
tsuyoshi38, line 71: It may be okay but do you need to delete the following line ?
!! SOURCE

AugustinLu: Reverted

src/md_control_module.f90
AugustinLu: Added in 9186b16

tsuyoshi38 (approved):

tsuyoshi38:
@AugustinLu  Please merge it.

PR #257, opened 2023-09-07
Fix issues with eigenvalue output format

davidbowler:
Add extra digits to k-point weight and number of k-point formatting in Conquest output.  Also ensure that projected DOS checks for limited energy window output from Conquest (read the appropriate flag)

### DIFF
--- a/src/io_module.f90
+++ b/src/io_module.f90
@@ -1641,7 +1641,7 @@ second:   do
     close(io)
     !    
 4   format(10x,'Sum of eigenvalues: ',f18.11,' ',a2)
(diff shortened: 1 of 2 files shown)


lionelalexandre (commented):
I might need to check the format of wtk for ase output as well

davidbowler:
From what I can see, you might need to change the format for the number of k-points to `i6` from `i3`:

https://github.com/OrderN/CONQUEST-release/blob/435fe322506705d89c3e813219d871aef9dcf753/src/io_module.f90#L1644

but the other issue (k-point weight) doesn't come up because you don't write them out

lionelalexandre (approved):

Issue #255, opened 2023-09-04
Output frequency for xsf and xyz trajectories


AugustinLu:
At present, when `AtomMove.WriteXSF = .TRUE.` and/or `AtomMove.WriteExtXYZ = .TRUE.`, trajectories in XSF and/or extXYZ format are written at each time step.

For long molecular dynamics simulations, these trajectory files can become huge, and not all frames are necessary.

Is it possible to add a keyword to control the frequency at which these files are written? Some possibilities could be:

1. Using `AtomMove.OutputFreq` to control all of them together. 
2. Defining `AtomMove.XsfFreq` and `AtomMove.XyzFreq` for control them separately. 

Thank you very much.

tsuyoshi38:
Personally, I prefer to have all of these keywords and the default values of `AtomMove.XsfFreq` and `AtomMove.XyzFreq` are set by `AtomMove.OutputFreq`.

AugustinLu:
An implementation of the two additional keywords is proposed in the forked repository: [4fa7fcf](https://github.com/AugustinLu/CONQUEST-release/commit/4fa7fcfc82bf4d275bbcba2eb16bb3d027fd775f)

- If no value is provided, the default value for `AtomMove.XsfFreq` and `AtomMove.XyzFreq` are equal to `AtomMove.OutputFreq`
- **trajectory.xsf** is written every  `AtomMove.XsfFreq` steps and **trajectory.xyz** is written every  `AtomMove.XyzFreq` steps, if the corresponding flags are set to `.TRUE.`.

davidbowler:
This looks like a very sensible solution: thank you.

davidbowler:
Closed by #258

PR #258, opened 2023-09-13
F xsf xyz write frequency

AugustinLu:
An implementation of the two additional keywords is proposed to control the frequency with which the XSF and XYZ coordinate files are written.

- If no value is provided, the default value for `AtomMove.XsfFreq` and `AtomMove.XyzFreq` are equal to `AtomMove.OutputFreq`
- **trajectory.xsf** is written every  `AtomMove.XsfFreq` steps and **trajectory.xyz** is written every  `AtomMove.XyzFreq` steps, if the corresponding flags are set to `.TRUE.`.


davidbowler (approved):

davidbowler:
I think that this is fine, so I have approved it and taken it out of draft status

src/initial_read_module.f90
```diff
@@ -1568,6 +1569,8 @@ subroutine read_input(start, start_L, titles, vary_mu,&
     CGreset               = fdf_boolean('AtomMove.ResetCG',          .false.)
     MDn_steps             = fdf_integer('AtomMove.NumSteps',     100        )
     MDfreq                = fdf_integer('AtomMove.OutputFreq',    50        )
+    XSFfreq               = fdf_integer('AtomMove.XsfFreq',    MDfreq        )
+    XYZfreq               = fdf_integer('AtomMove.XyzFreq',    MDfreq        )
```
JianboLin, line 1573: Some minority considerations, 
How about make `AtomMove.XsfFreq` and` AtomMove.XyzFreq` to `AtomMove.OutputFreqXSF` and` AtomMove.OutputFreqXYZ`? As sub option of `AtomMove.OutputFreq` and also consistent to `AtomMove.WriteXSF` and `AtomMove.WriteExtXYZ`.
Is it okay to keep default value to '1', since it is usually considered to print out 'XYZ' or 'XSF' each step in common DFT calculations?

AugustinLu: >How about make AtomMove.XsfFreq and AtomMove.XyzFreq to AtomMove.OutputFreqXSF and AtomMove.OutputFreqXYZ? As sub option of AtomMove.OutputFreq and also consistent to AtomMove.WriteXSF and AtomMove.WriteExtXYZ.

Strictly speaking, they are not a sub option of `AtomMove.OutputFreq`, since they could be written independently from the other files (`md.frames`, etc.). 

>Is it okay to keep default value to '1', since it is usually considered to print out 'XYZ' or 'XSF' each step in common DFT calculations?

I am not sure what is the best option. (see also https://github.com/OrderN/CONQUEST-release/issues/255#issuecomment-1704625057).  For MD in general, it is not common to write every step. For geometry relaxation (with cg, sqnm, etc.), writing every step is probably useful.

PR #260, opened 2023-09-13
Read distance units from input file

davidbowler:
This is needed to correctly convert units for simulation cell and coordinates in io_module.f90 and wasn't set before.


tsuyoshi38 (approved):

tsuyoshi38:
`BohrToAng` is used if the number in the input file is in Angstrom unit, and changed to Bohr unit, I think. Am I correct ?  Then, it sounds me more appropriate to use `AngToBohr` .   Am I misunderstanding something ?

davidbowler:
Thank you for the comment: it is confusing! The variable `dist_conv` is used to convert internal Conquest distances (in Bohr) to output units selected by the user.  I've added a comment to explain this.

PR #262, opened 2023-09-14
Add input configurations used for profiling

### LINKED ISSUE #261: Store inputs for performance tests
It would be useful to keep inputs to the performance tests we run under version control in the repository, so that they are easier to reference and share, and reuse in later projects. Keeping them under the testsuite is not ideal, at least without some changes to the python code that checks the outputs in the CI workflow. I would create a separate subdirectory in the main repository.

ilectra:
Fixes #261 
These are systems we run for profiling, benchmarking, or scaling studies, thus they're too large to be run as CI tests.


tkoskela (changes requested):
Could you change the base branch to `develop`?

tkoskela (approved):

ilectra:
@davidbowler could you please add a few words in a README for every benchmark? Otherwise we are ready to merge this.

ilectra:
> @davidbowler could you please add a few words in a README for every benchmark? Otherwise we are ready to merge this.

Moved it to #275 , we need to merge this PR.

PR #267, opened 2023-09-21
Set omp_stacksize in testsuite

### LINKED ISSUE #254: OMP_STACKSIZE should be set for testsuite
Running test003 with more than one thread (on at least one platform) fails with stack overflow error unless OMP_STACKSIZE is set to be large enough (simple tests suggest 50M is enough).

It should be simple to fix: I assume that we can set a stack size in the appropriate places (the bash script `run_conquest_tests.sh` and the GitHub integration)

tkoskela:
Closes #254 

I set `OMP_STACKSIZE` to 100M as a first guess, let's see how that performs.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -15,6 +15,7 @@ jobs:
       F77: mpif77
       OMP_NUM_THREADS: ${{matrix.threads}}
       OPENBLAS_NUM_THREADS: 1
+      OMP_STACKSIZE: 100M
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


ilectra:
Have we documented this somewhere? (that the user might need to change the OMP_STACKSIZE) I vaguely remember that we did, but I cannot find it now. Otherwise, it's fine to merge.

ilectra (approved):

davidbowler:
This is documented under Installation in the manual (though only the develop branch, as that's as far as this has got)

davidbowler (approved):

PR #272, opened 2023-09-22
Version Control system.make for different machines

### LINKED ISSUE #247: Version control system.make for different machines
At the moment, the user needs to make local changes to `system.make` to build on different systems with different compilers, library paths, etc. This causes a lot of faff with git not being able to deal with local changes. My suggestion is: keep version controlled versions of `system.make` for machines we commonly run on, perhaps in a separate subdirectory. Modify the makefile so that it attempts to detect the machine it is building on (e.g. with `hostname -d`, or something similar) and include the correct `system.make` for that machine.

ilectra:
Fixes #247


tkoskela:
It seems that the GitHub actions runners give you a hostname that looks difficult to predict. 

```
Running on SYSTEM .2dmmb0nmsnvutc1liyrbff0gyd, using makefile system/system.2dmmb0nmsnvutc1liyrbff0gyd.make
```

Maybe we could check for a plain `system.make` if the hostname'd one is not found? (or possibly check for it first)

ilectra:
> It seems that the GitHub actions runners give you a hostname that looks difficult to predict.
> 
> ```
> Running on SYSTEM .2dmmb0nmsnvutc1liyrbff0gyd, using makefile system/system.2dmmb0nmsnvutc1liyrbff0gyd.make
> ```
> 
> Maybe we could check for a plain `system.make` if the hostname'd one is not found? (or possibly check for it first)

I made the `gha.make` be copied to the correct `system.xxxxx.make`. But what you suggest also makes sense.... Could do either.

Now, why is the CI not running again in my last commit????....

ilectra:
I think this is ready, unless you want to add your `system.thomas.make` file @tkoskela

ilectra:
@tkoskela done, please review!

src/system/system.example.make
```diff
@@ -41,6 +39,14 @@ XC_COMPFLAGS = -I/usr/local/include
 FFT_LIB=-lfftw3
 FFT_OBJ=fft_fftw3.o
 
+# Full library call; remove -lscalapack if using dummy diag module.
+# If using OpenMPI, use -lscalapack-openmpi instead.
+LIBS= $(FFT_LIB) $(XC_LIB) -lscalapack $(BLAS)
```
tkoskela, line 44: You fixed `LIBS` being set before these are defined, that's good!

docs/installing.rst
```diff
@@ -44,8 +44,13 @@ a working MPI installation including a Fortran90 compiler (often
...
 
-The library locations are set in the ``system.make`` file in the ``src/``
-directory, along with other parameters needed for compilation.
+The library locations are set in the ``system.<systemname>.make`` file in the ``src/system``
+directory, along with other parameters needed for compilation. ``system.<systemname>.make``
+files are provided for some HPC systems used by the community, but if you want to run
+locally or on a different system, you need to provide an appropriate ``system.<systemname>.make``
+file. Use ``src/system/system.example.make`` as a starting point. If you are running on
+a local machine (``hostname -d`` returns empty), the system-specific makefile should be named
+``system.make``.
```
tkoskela, line 53: Should this also include documentation on how to add new `system.*.make` files?

ie. check `hostname -d`, name your file appropriately and copy it to `system/`

tkoskela (commented):
On young I have 
```
$ hostname -d
ib.young.ucl.ac.uk
```
so that 
```
$ hostname -d | cut -d. -f1
ib
```
while I can use `ib` as the label for the `system.make` file. Could we make the parsing more robust?

src/Makefile
```diff
@@ -24,7 +24,18 @@ MULT_KERN = default
...
+SYSTEM = $(shell hostname -d | cut -d. -f1)
+ifneq ($(SYSTEM),)
+SYSTEM := .$(SYSTEM)
+endif
+SYSTEM_PATH = system/system$(SYSTEM).make
+ifneq ("$(wildcard $(SYSTEM_PATH))","")
+$(info Building on SYSTEM $(SYSTEM), using makefile $(SYSTEM_PATH))
+include $(SYSTEM_PATH)
+else
+$(error Cannot find $(SYSTEM_PATH) file for system $(SYSTEM). Please make one, using system/system.example.make as an example)
```
tkoskela, line 37: This is pretty minor, but since you've prefixed `SYSTEM` with a `.` it gets printed as `.<system name>` here

src/Makefile
```diff
@@ -24,7 +24,18 @@ MULT_KERN = default
...
 #Include system-dependent variables
-include system.make
+#Get the system name automatically.
+SYSTEM = $(shell hostname -d | cut -d. -f1)
+ifneq ($(SYSTEM),)
+SYSTEM := .$(SYSTEM)
+endif
+SYSTEM_PATH = system/system$(SYSTEM).make
+ifneq ("$(wildcard $(SYSTEM_PATH))","")
+$(info Building on SYSTEM $(SYSTEM), using makefile $(SYSTEM_PATH))
```
tkoskela, line 34: This is pretty minor, but since you've prefixed `SYSTEM` with a `.` it gets printed as `.<system name>` here

src/system/system.kathleen.make
```diff
@@ -0,0 +1,59 @@
+#
```
tkoskela, line 1: There was a suggestion from @davidbowler to add some info about the system in the header

ilectra:
@tkoskela this should be ready for re-review/merging.

tkoskela (approved):

PR #285, opened 2023-11-08
fix hard-coded scalapack link flag

tkoskela:


davidbowler:
We should have this argument associated with the `DIAG_DUMMY` variable as well - they are selecting the same behaviour

ilectra (approved):

tkoskela:
I'm going to merge this because it will help build the spack package more smoothly. The change was not really about selecting whether we build with or without scalapack, just putting the flag somewhere where it's easier to change programmatically (because of `cray-libsci` shenanigans). We can think about further automations to the build systems later.

PR #290, opened 2023-11-27
Overlap communication with computation in multiply_module

### LINKED ISSUE #265: Overlap communication with computation in multiply_module
Things to look out for
- Can we overlap communication and calculation in https://github.com/OrderN/CONQUEST-release/blob/6bf8f4a8c20fd4fa8f1c7baeb8a6b1f23a6d2408/src/multiply_module.f90#L251

_Originally posted by @tkoskela in https://github.com/OrderN/CONQUEST-release/issues/248#issuecomment-1697085794_
            
- Can we use OpenMP tasks to both receive data and call multiply kernels, with appropriate dependencies

ilectra:
Fixes #265

### DIFF
--- a/src/multiply_module.f90
+++ b/src/multiply_module.f90
@@ -155,13 +155,10 @@ contains
     integer(integ),pointer :: ibseq_rem(:)
     integer(integ),pointer :: ibind_rem(:)
     integer(integ),pointer :: ib_nd_acc_rem(:)
-    integer(integ),pointer :: npxyz_rem(:)
     integer(integ),pointer :: ibndimj_rem(:)
     ! Arrays for remote variables to point to
-    integer, target :: part_array(3*a_b_c%parts%mx_mem_grp+ &
-         5*a_b_c%parts%mx_mem_grp*a_b_c%bmat(1)%mx_abs)
     integer, dimension(:), allocatable :: nreqs
-    integer :: offset,sends,i,j
+    integer :: sends,i,j
     integer, dimension(MPI_STATUS_SIZE) :: mpi_stat
     integer, allocatable, dimension(:) :: recv_part
     real(double) :: t0,t1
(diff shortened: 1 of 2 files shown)


tkoskela:
From @davidbowler: compute is only being called on kpart [2 : end]. To fix, call compute kernels on kpart -1 then call once after the loop on kpart.

ilectra:
I think this can be reviewed now. I'll produce some profiles, to see if we gained anything, when I'm back from the holidays, I don't think I'll have time tomorrow.

src/multiply_module.f90
```diff
@@ -567,40 +675,56 @@ subroutine prefetch(this_part,ahalo,a_b_c,bmat,icall,&
...
     real(double), dimension(lenb_rem) :: b_rem
     real(double) :: b(:)
     ! Local variables
     integer :: ncover_yz,ind_part,iskip,ind_last
     integer :: inode,ipart,nnode
+    logical :: do_nonb_local
 
+    ! Set non-blocking receive flag
+    do_nonb_local = .false.
+    if (present(do_nonb)) do_nonb_local = do_nonb
```
tkoskela, line 695: Could we remove the `do_nonb` argument, and just check for `if(present(request))` here? Then you wouldn't have to check for that later? Is there a benefit of being able to handle a scenario where `do_nonb = .false.` gets passed in?

tkoskela (commented):
I did a pass of reviews, sorry about being a bit nitpicky in some places. It's Friday.

I would like to understand the performance better before we merge this in.

ilectra:
There's no performance improvement seen, if anything there's a small degradation (see below). I think I understand why: the problem is the order communications are received, and not the time they take.
Screenshot from 2024-06-17 12-44-45
Will not merge for now, and instead investigate optimising the order in https://github.com/OrderN/CONQUEST-release/tree/ic-mm-comms-optimise-order . If that works, then we can revisit overlapping comms with computation, for further improvement.

PR #291, opened 2023-11-29
Test multiply kernels in CI

### LINKED ISSUE #270: Test different multiply kernels in testsuite
While working on the threading implementations of the matrix multiply kernel, I realised it would be very useful to test more of them than just the default. Unless we can converge on a single optimal implementation I think we should test at least one threaded and one non-threaded implementation, perhaps also a gemm and a non-gemm one. 

It can be fairly easily implemented by adding the multiply kernel into the test matrix. If we use spack to automate the build it's just a matter of changing the variant in the spack spec. I'm a bit concious of blowing up the number of jobs in the test matrix, let's discuss this in the next meeting.

tkoskela:
Closes #270

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -64,6 +64,7 @@ jobs:
       
     - name: Spack install
       run: |
+        source /opt/spack/share/spack/setup-env.sh
         spack --version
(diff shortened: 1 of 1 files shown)


tkoskela:
Running spack in the CI workflow was not a good idea

PR #293, opened 2023-11-30
Fix multiply kernel memory leak

### LINKED ISSUE #280: Memory leak for linear scaling ompGemm_m kernel on atom movement
I have identified a significant memory leak in O(N) geometry optimisation, though not the source.  The system (which I will attach input files for when I have a better internet connection) is 4,096 atoms of silicon (8x8x8 fundamental cells) running on 64 processes (though this does not seem to matter) with a SZ basis and L range of 16 bohr.  With the develop branch, at some point after the initial movement of atoms the memory use spikes (I've seen at least 16GB in one process) and even with 1.5TB of RAM the job fails.  The same behaviour is not seen for v1.2.

tkoskela:
Closes #280 

Uses assumed shape arrays for remote index arrays. These can vary in size and using a fixed size leads to accessing uninitialized memory.

Edit: I wanted to clarify that the inputs you gave in #292 still fail for me on all multiply kernels. So I'm not claiming to fix *that* issue. But I discovered a different issue and I'm fixing it :grinning: 

Edit2: The reason the inputs from #292 fail for me is that my machine runs out of memory. Based on our last discussion that is to be expected with the size of the problem.

### DIFF
--- a/src/matrix_module.f90
+++ b/src/matrix_module.f90
@@ -673,6 +673,8 @@ contains
     if(stat/=0) then
        call cq_abort('alloc_halo: error allocating ndimi')
     endif
+    halo%ndimi = 0
+    halo%ndimj = 0
     call stop_timer(tmr_std_allocation)
     return
   end subroutine allocate_halo
--- a/src/multiply_kernel_default.f90
+++ b/src/multiply_kernel_default.f90
@@ -148,12 +148,12 @@ contains
     real(double)       :: c(lenc)
     integer, optional  :: debug
     ! Remote indices
-    integer(integ) :: ib_nd_acc(mx_part)
-    integer(integ) :: ibaddr(mx_part)
-    integer(integ) :: nbnab(mx_part)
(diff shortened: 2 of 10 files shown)


tkoskela:
I tested this change on all multiply kernels in https://github.com/OrderN/CONQUEST-release/actions/runs/7045441423 and the tests pass so it's ready

ilectra (approved):
Looks fine, but I still don't understand why it worked! See my review comment in https://github.com/OrderN/CONQUEST-release/pull/292#discussion_r1410570980

tkoskela:
Actually, with this change we can just pass slices of `part_array` into the multiply kernel, there is no need for pointers (other than making the subroutine calls look a bit more tidy). I would definitely recommend

davidbowler:
I have added two small but important changes: we now zero `part_array` and `ndimi` and `ndimj` in the halo structure, all of which are passed to `maxval`.  This should now fix the problem in #280

tsuyoshi38 (approved):

PR #294, opened 2023-11-30
Remove pointers in multiply module

tkoskela:
This PR removes pointers to `part_array` in `multiply_module` and passes slices of `part_array` to subroutines instead. It also tries to clarify the indexing of `part_array` to make it clearer which slice is being passed.

I'm open to opinions on whether or not this is a good idea. It's more of a style change than anything else, although it should make the code more robust against the kind of memory errors in #280. I would argue this is a more Fortran style use of arrays, compared to the C-style use of pointers. It does make the subroutine calls somewhat longer

### DIFF
--- a/src/matrix_comms_module.f90
+++ b/src/matrix_comms_module.f90
@@ -75,10 +75,10 @@ contains
     ! Passed variables
     integer :: myid
     integer :: n_cont,ncover_yz,ncoverz
-    integer(integ) :: nbnab_rem(:)
-    integer(integ) :: ibind_rem(:)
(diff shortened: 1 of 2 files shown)


ilectra (approved):

ilectra:
Actually, can this PR please wait until we merge #290 @tkoskela ? The conflicts will be a much bigger pain if we do it the other way round....

davidbowler:
I want to think about this change some more, and discuss it with @tsuyoshi38 before we approve it

davidbowler (approved):
This is a better way to manage the access to `part_array`

PR #297, opened 2023-12-20
Add ID strings to matrix multiplication kernels and write out

davidbowler:
This enables us to see which matrix multiplication kernel is being used; given that the code changes with kernel, I think that this record is important.


ilectra (approved):

tkoskela (approved):
Looks good to me!

tsuyoshi38 (approved):

PR #302, opened 2024-01-11
Compile pseudo_tm_info.f90 without optimisation

davidbowler:
This seems to fix a strange issue with GCC13 on Mac


tsuyoshi38 (approved):
I am not sure whether we should change Makefile only for this version of Mac..  But, it is okay if temporary.

tkoskela:
Do we know where in `pseudo_tm_info.f90` this issue occurs?

PR #303, opened 2024-01-12
Fix order of flags in Makefile

tkoskela:
While trying to compile without the mpi compiler wrapper, I noticed that the order of linker flags in the Makefile seems incorrect. According to [this stackoverflow thread](https://stackoverflow.com/questions/35897290/what-is-the-right-order-of-linker-flags-in-gcc) the reverse order used to work but ceased a while ago.With the order of flags in current `develop`, I would fail to link the MPI library with `gfortran-13` after adding the flags given by `mpifort --showme:link` to `LINKFLAGS` in my `system.make`. Changing to the correct the order fixed my issue.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -77,7 +77,7 @@ deps.obj.inc: $(SRCS_NODS) $(SYSTEM_PATH)
 
 #Target
 $(TARGET) : $(NODE_OBJECTS)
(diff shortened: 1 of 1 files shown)


davidbowler (approved):
Thank you for catching this

PR #307, opened 2024-01-18
Change distance criteria to be consistent

davidbowler:
In finding matrix elements, we now check for r2 < rcut2 - RD_ERR in all routines (before we sometimes used a locally defined tolerance, and in some places we had + instead of -)

I think that this was the source of the error (my guess is that the change in `set_bucket_module.f90` was the key one when compared to `matrix_elements_module.f90` but I'm not sure).

### DIFF
--- a/src/matrix_elements_module.f90
+++ b/src/matrix_elements_module.f90
@@ -259,6 +259,7 @@ contains
     ! Module usage
     use datatypes
     use basic_types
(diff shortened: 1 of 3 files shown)


tsuyoshi38 (approved):
Thank you for this debugging. Now, we decide that we don't consider the pairs of atoms or atom-grid as neighbors, if the distance is exactly same as the cutoff.

PR #316, opened 2024-02-02
Corrected the expression for the temperature change rate (control.f90).

### LINKED ISSUE #315: Temperature change per step in variable temperature MD
In variable temperature MD, the change of temperature at each step is wrongly calculated as the ratio between the cooling/heating rate and the time step, instead of the product. Therefore, the change per time step is wrong if the time step is not 1 fs.

`temp_change_step = md_variable_temperature_rate / mdl%timestep`

should be

`temp_change_step = md_variable_temperature_rate * mdl%timestep`

Sorry for the inconvenience.

AugustinLu:
Closes #315 

Correct definition of the change of temperature at each step in variable temperature MD simulations.

### DIFF
--- a/src/control.f90
+++ b/src/control.f90
@@ -854,7 +854,7 @@ contains
          ! At a given time step, update T_ext and ke_target
          ! Temperature evolves linearly from md_initial_temperature to md_final_temperature by step of temp_change_step
          ! Stops when target temperature has been reached (i.e. abs(dT) < abs(temp_change_step) )
-         temp_change_step = md_variable_temperature_rate / mdl%timestep ! Unit is K
+         temp_change_step = md_variable_temperature_rate * mdl%timestep ! Unit is K
          mdl%T_ext = mdl%T_ext + temp_change_step
          thermo%ke_target = half*md_ndof_ions*fac_Kelvin2Hartree*mdl%T_ext ! Update target ke for SVR
 


davidbowler (approved):
We should think about units for the variable temperature rate, maybe (I guess they will be K/fs by default: is that made clear anywhere?)

tsuyoshi38 (commented):
I am sorry for asking the following now, but ..

Line 845-850:   "exit" instead of  "stop"?
If you want to stop here, please use "cq_abort".
If just "exit", calling "cq_warn" is better.

AugustinLu:
> I am sorry for asking the following now, but ..
> 
> Line 845-850: "exit" instead of "stop"? If you want to stop here, please use "cq_abort". If just "exit", calling "cq_warn" is better.

The idea is that it should be a "failure" of the calculation and nothing should be calculated at all.
Therefore, `exit` should be replaced by a call to `cq_warn`, right?

Just to confirm, `exit` is the proper termination of CONQUEST -> Is it correct?

tsuyoshi38:
Then, please use `cq_abort`.  
If you use `exit` here, it would just exit the do-loop, I think.

AugustinLu:
> Then, please use `cq_abort`. If you use `exit` here, it would just exit the do-loop, I think.

You are right. I will update this.

Issue #305, opened 2024-01-18
Add extended XYZ file output for static run and structural optimization


JianboLin:
We have an option for output of extended-xyz file in run types of MD simulation.
But when we would like to check, save, or even visualize the atomic forces of specific geometry or during the structural optimization, add the same option in run types of static and structural optimization, would be convenient.

I plan to do this with following workflow:
1. Add to and check in run types of static and structural optimization(cg_run, lbfgs, and sqnm),
2. Add to and check in run types of structural optimization with cell optimization.

JianboLin:
I have checked this implementation with static, cg, lbfgs, sqnm and also with cell optimization.
All of them can output extended xyz files with
`AtomMove.WriteExtXYZ T`

The following shows a test with Si  8 atoms, cubic cell (shifted first atom in x axis) by static run:
Coordination file(length unit: Bohr):
```
10.262912871373274 0.0000000000000 0.0000000000000
0.0000000000000 10.262912871373274 0.0000000000000
0.0000000000000 0.0000000000000 10.262912871373274
8
0.0500000000000 0.0000000000000 0.0000000000000 1 T T T
0.5000000000000 0.5000000000000 0.0000000000000 1 T T T
0.5000000000000 0.0000000000000 0.5000000000000 1 T T T
0.0000000000000 0.5000000000000 0.5000000000000 1 T T T
0.2500000000000 0.2500000000000 0.2500000000000 1 T T T
0.7500000000000 0.7500000000000 0.2500000000000 1 T T T
0.7500000000000 0.2500000000000 0.7500000000000 1 T T T
0.2500000000000 0.7500000000000 0.7500000000000 1 T T T
```

In Conquest_out, we have:
```
    force: Forces on atoms (Ha/a0)
    force:  Atom   X              Y              Z
    force:     1  -0.0915292599  -0.0000000000  -0.0000000000
    force:     2   0.0051225981  -0.0000000000  -0.0000000000
    force:     3   0.0051225981   0.0000000000  -0.0000000000
    force:     4  -0.0083033384  -0.0000000000   0.0000000000
    force:     5   0.0245641018   0.0235894111   0.0235894111
    force:     6   0.0201767060   0.0136070745  -0.0136070745
    force:     7   0.0201767060  -0.0136070745   0.0136070745
    force:     8   0.0245641018  -0.0235894111  -0.0235894111
```

In trajectory.xyz, we get :
```
8
[...]

PR #318, opened 2024-02-06
Add extended XYZ file output for static run and structural optimization

JianboLin:
We have an option for output of extended-xyz file in run types of MD simulation.
But when we would like to check, save, or even visualize the atomic forces of specific geometry or during the structural optimization, add the same option in run types of static and structural optimization, would be convenient.

At present, from my tests, it works with static, cg, lbfgs, sqnm and also with cell optimization
detail in https://github.com/OrderN/CONQUEST-release/issues/305

### DIFF
--- a/src/control.f90
+++ b/src/control.f90
@@ -496,7 +496,7 @@ contains
        end if
        call write_atomic_positions("UpdatedAtoms.dat", trim(pdb_template))
        if (flag_write_xsf) call write_xsf('trajectory.xsf', iter)
-       if (flag_write_extxyz) call write_extxyz('trajectory.xyz', energy1, tot_force)
+       if (flag_write_extxyz .and. mod(iter,XYZfreq) == 0) call write_extxyz('trajectory.xyz', energy1, tot_force)
        ! Analyse forces
        g0 = dot(length, tot_force, 1, tot_force, 1)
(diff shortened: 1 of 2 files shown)


davidbowler (commented):
We have a variable `XYZfreq` (in `control.f90`) which is used to define how often this file is written out (at present only during MD, in `md_misc_module.f90`).  I wonder whether we should use it here: what do you think?

JianboLin:
Thank you for your comment.
Static calculation or structural optimization (including opt cell ) may not have large number of steps, supposing several hundred steps.
And it may be also okay if we use default `XYZfreq=1`, in these run types (except static).

tsuyoshi38:
> Thank you for your comment. Static calculation or structural optimization (including opt cell ) may not have large number of steps, supposing several hundred steps. 

I understand your point.

> And it may be also okay if we use default `XYZfreq=1`, in these run types (except static).

But, since it might help in some cases, could you do it?

JianboLin:
Yes, updated in the latest commit. By the way, do we also to control output frequency of xsf file with `XSFfreq` in these types of run?

tsuyoshi38 (approved):

JianboLin:
Thank you for your approval.

PR #319, opened 2024-02-06
Properly abort the calculation if wrong method for variable temperatu…

AugustinLu:
Related to #316 

> Then, please use cq_abort.
If you use exit here, it would just exit the do-loop, I think.

When a wrong method for variable temperature MD is set, the program should abort and not just exit the loop.

Solution: Only the `linear` method is accepted. Any other value will lead to error message and CONQUEST will abort:

image

### DIFF
--- a/src/initial_read_module.f90
+++ b/src/initial_read_module.f90
@@ -2299,6 +2299,19 @@ contains
     ! Variable temperature
     flag_variable_temperature = fdf_boolean('MD.VariableTemperature', .false.)
     md_variable_temperature_method = fdf_string(20, 'MD.VariableTemperatureMethod', 'linear')
+
+    ! Verify that a method for variable temperature is valid
+    if (flag_variable_temperature) then
+      ! At present, only linear evolution is supported
+      if (md_variable_temperature_method .ne. 'linear') then
+        if(inode==ionode) then
+          write(io_lun,fmt='(6x, "Wrong method for variable temperature: ", a, " != linear. Stopping ..." )') &
+            trim(md_variable_temperature_method)
+        end if
+        call cq_abort("Wrong method for variable temperature")
+      end if
+    end if
+
     md_variable_temperature_rate = fdf_double('MD.VariableTemperatureRate', 0.0_double)
     md_initial_temperature = fdf_double('MD.InitialTemperature',temp_ion)
     md_final_temperature = fdf_double('MD.FinalTemperature',temp_ion)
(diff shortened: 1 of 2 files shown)


davidbowler (changes requested):
The output should be written to the `io_lun` unit, not to `*` so that it is recorded in the output file.  But I wonder if this check should be done in `read_input` in `initial_read_module.f90` when we're reading in the various parameters: it would make sense to stop much earlier, I think.

AugustinLu:
> The output should be written to the `io_lun` unit, not to `*` so that it is recorded in the output file. 

It is indeed better to have in the output file rather than standard output. I will change it.

>But I wonder if this check should be done in `read_input` in `initial_read_module.f90` when we're reading in the various parameters: it would make sense to stop much earlier, I think.

Oh, yes. This is indeed the way it should be done. I will move that check to just after reading the input parameters, so that no time is wasted in initialising the calculation for nothing.

AugustinLu:
In commit 0cf34260115ffc55ddc1d17beecbbfec444259bb, the target for the output has been updated from `*` to `io_lun` and the block for the verification of the variable temperature method has been moved from `control.f90` to `initial_read_module.f90`.

Contents of output file:
image

Standard output:
image

src/initial_read_module.f90
```diff
@@ -2299,6 +2299,19 @@ subroutine read_input(start, start_L, titles, vary_mu,&
     ! Variable temperature
     flag_variable_temperature = fdf_boolean('MD.VariableTemperature', .false.)
     md_variable_temperature_method = fdf_string(20, 'MD.VariableTemperatureMethod', 'linear')
+
+    ! Verify that a method for variable temperature is valid
+    if (flag_variable_temperature) then
+      ! At present, only linear evolution is supported
+      if (md_variable_temperature_method .ne. 'linear') then
```
davidbowler, line 2306: Can we please use `leqi` from `input_module` to compare strings? Something like `if(.not.leqi(md_variable_temperature_method(1:6),'linear')` would work I think and be more portable (in particular it will ignore case...)

davidbowler (approved):

AugustinLu:
> Can we please use `leqi` from `input_module` to compare strings? Something like `if(.not.leqi(md_variable_temperature_method(1:6),'linear')` would work I think and be more portable (in particular it will ignore case...)

Thank you for the suggestion. 
I have updated and checked the code accordingly.

tsuyoshi38 (approved):

Issue #308, opened 2024-01-19
Kathleen system.make file is using Conquest XC library


tkoskela:
It should use the system installed XC library version 6.2.2 instead

davidbowler:
Since this is a specific file for a specific system, I guess that this is reasonable (i.e. we know LibXC is available on Kathleen) though in general the Conquest XC is the safest default (it will work on all systems).

tkoskela:
Closed via #331

PR #320, opened 2024-02-06
276 combine nsf loops

connoraird:
## Description
- We have combined three consecutive loops over support functions (nsup) into a single one. 
- This loop has then been threaded and collapsed 
    - To allow this, some memory allocation had to be refactored and moved outside of the parallel region. 

## Speedup plot
This plot shows the performance of test `test_004_isol_C2H4_4proc_PBE0CRI` for 1 mpi process
276-combine-nsf-loops


src/exx_kernel_default.f90
```diff
@@ -1173,36 +1151,25 @@ subroutine m_kern_exx_cri(k_off, kpart, ib_nd_acc, ibaddr, nbnab, &
...
-                   if ( exx_alloc ) call exx_mem_alloc(extent,jb%nsup,0,'phi_j','dealloc')                   
-                   if ( exx_alloc ) call exx_mem_alloc(extent,kg%nsup,jb%nsup,'vhf_kj','alloc')                   
+                   if ( exx_alloc ) call exx_mem_alloc(extent,0,0,'Ome_kj','alloc')
                    !
+                   call start_timer(tmr_std_exx_accumul)
+                   !$omp parallel do schedule(runtime) collapse(2) default(none) reduction(+: c) &
+                   !$omp     shared(kg,jb,tmr_std_exx_poisson,tmr_std_exx_accumul,Phy_k,phi_j,phi_k,ncbeg,ia,tmr_std_exx_matmult,ewald_pot,phi_i, &
+                   !$omp            exx_psolver,exx_pscheme,extent,dv,ewald_rho,inode,pulay_radius,p_omega,p_gauss,w_gauss,reckernel_3d,r_int) &
+                   !$omp     private(nsf1,nsf2,work_out_3d,work_in_3d,ewald_charge,Ome_kj, &
+                   !$omp             ncaddr,nsf3,exx_mat_elem,r,s,t)
```
tkoskela, line 1166: Could you shorten the `shared()` lines? Sticking to the 132 character limit would be good style. Otherwise we need to start enforcing the no line length limit in the compiler flags.

src/exx_kernel_default.f90
```diff
@@ -1173,36 +1151,25 @@ subroutine m_kern_exx_cri(k_off, kpart, ib_nd_acc, ibaddr, nbnab, &
...
                    do nsf1 = 1, kg%nsup
                       do nsf2 = 1, jb%nsup
 
-                         call start_timer(tmr_std_exx_poisson)    
-                         work_in_3d  = zero
                          work_out_3d = zero
                          !
-                         work_in_3d  = rho_kj(:,:,:,nsf1,nsf2)
+                         work_in_3d = Phy_k(:,:,:,nsf1) * phi_j(:,:,:,nsf2)
```
tkoskela, line 1172: Do we really need these temporary buffers? I understand conceptually you want to store temporary results in a separate buffer. From a coding point of view, couldn't you just pass `Phy_k` and `phi_j` into `exx_ewald_charge` and use `Ome_kj` to store the temporary results?

`work_out_3d` is also being zeroed twice, and then overwritten.

src/system/system.kathleen.ucl.ac.uk.make
```diff
@@ -13,7 +13,7 @@ OMPFLAGS= -fopenmp
 
 # Compilation flags
 # NB for gcc10 you need to add -fallow-argument-mismatch
-COMPFLAGS= -O3 -g $(OMPFLAGS) $(XC_COMPFLAGS) -I"${MKLROOT}/include"
+COMPFLAGS= -xAVX -O3 -g $(OMPFLAGS) $(XC_COMPFLAGS) -I"${MKLROOT}/include"
```
tkoskela, line 16: Could you move the modifications to `system.kathleen.*.make` into a separate PR that closes #308 ?

It doesn't make much difference here but in general using `-xHost` is better than `-xAVX`. `-xHost` will select the highest level of vectorization supported by the hardware. Yes I know I told you to use `-xAVX`, it's my fault :see_no_evil:

src/exx_kernel_default.f90
davidbowler: work_in_3d is altered below (l.1176)

lionelalexandre (approved):

connoraird:
Results of running bench marks 
[CONQUEST 276 parallel speedup.pdf](https://github.com/OrderN/CONQUEST-release/files/14360611/CONQUEST.276.parallel.speedup.pdf)

src/exx_memory.f90
```diff
@@ -107,6 +106,15 @@ subroutine exx_mem_alloc(extent,nsf1,nsf2,matrix,flag,unit,n_neigh,neigh)
...
           !
+       case('phi_i_1d_buffer') ! allocate phi_i_1d_buffer for primary atom
+          allocate(phi_i_1d_buffer(nsf1*(2*extent+1)*(2*extent+1)*(2*extent+1)), STAT=stat)
+          if(stat/=0) call cq_abort('Error allocating memory to phi_i_1d_buffer/exx !',stat)
+          call reg_alloc_mem(area_exx,nsf1*(2*extent+1)*(2*extent+1)*(2*extent+1),&
+               type_dbl,matrix,lun)
+          phi_i_1d_buffer = zero
+          !write(*,*) '\phi_{i\alpha}_1d_buffer allocated', shape(phi_i_1d_buffer)
+          !
+          !
```
tkoskela, line 117:
```suggestion
```

src/exx_memory.f90
```diff
@@ -107,6 +106,15 @@ subroutine exx_mem_alloc(extent,nsf1,nsf2,matrix,flag,unit,n_neigh,neigh)
           !write(*,*) '\phi_{i\alpha} allocated', shape(phi_i)
           !
           !
```
tkoskela, line 108:
```suggestion
```

tkoskela (changes requested):
This looks good to me. Only code changes I suggested is tidying up some commented out code (that probably was there before).

I would like some documentation of what you did, before we all forget about it. Could you either in the main comment of this PR, or in comments in the code (or both), briefly describe the main changes and the reasoning behind them?

tkoskela (approved):

lionelalexandre (approved):

PR #323, opened 2024-02-09
276 use blas

connoraird:
## Description
- Updates the rst nested loop in the kernel to use a BLAS dot product
    - This required refactoring some allocatable arrays into 1d buffers and using pointer arrays to reference them.
- Some additional unused allocatable arrays were also removed

## Speedup plot
This plot shows the performance of test `test_004_isol_C2H4_4proc_PBE0CRI` for 1 mpi process
276-use-blas

### DIFF
--- a/src/exx_kernel_default.f90
+++ b/src/exx_kernel_default.f90
@@ -936,7 +936,7 @@ contains
          exx_psolver, exx_pscheme, &         
          unit_exx_debug
     !
-    use exx_types, only: phi_i, phi_j, phi_k, phi_l, &
+    use exx_types, only: phi_i_1d_buffer, phi_j, phi_k, phi_l, &
          Phy_k, Ome_kj_1d_buffer, &
          work_in_3d, work_out_3d
     use exx_types, only: exx_alloc
--- a/src/exx_kernel_default.f90
+++ b/src/exx_kernel_default.f90
@@ -1199,22 +1199,6 @@ contains
                          !
                          do nsf3 = 1, ia%nsup
                             !
-                           !  exx_mat_elem = zero
-                           !  !
-                           !  do r = 1, 2*extent+1
-                           !     do s = 1, 2*extent+1
-                           !        do t = 1, 2*extent+1                         
-
-                           !           exx_mat_elem = exx_mat_elem &                                    
-                           !                + phi_i(t,s,r,nsf3)    &
-                           !                * Ome_kj_reduced(t,s,r) * dv
-
-                           !        end do
-                           !     end do
-                           !  end do
-                           !  !
-                           !  c(ncaddr + nsf3 - 1) = c(ncaddr + nsf3 - 1) + exx_mat_elem
-
                             c(ncaddr + nsf3 - 1) = c(ncaddr + nsf3 - 1) + dot((2*extent+1)**3, phi_i(:,:,:,nsf3), 1, Ome_kj, 1) * dv
                             !
                          end do ! nsf3 = 1, ia%nsup
--- a/src/exx_kernel_default.f90
+++ b/src/exx_kernel_default.f90
@@ -1122,7 +1122,7 @@ contains
           if ( exx_alloc ) call exx_mem_alloc(extent,ia%nsup,0,'phi_i_1d_buffer','alloc')
           print *, "Allocated phi_i_1d_buffer"
           phi_i(1:2*extent+1, 1:2*extent+1, 1:2*extent+1, 1:ia%nsup) => phi_i_1d_buffer
-          print *, "pointed phi_i at phi_i_1d_buffer"^
+          print *, "pointed phi_i at phi_i_1d_buffer"
           !
           call exx_phi_on_grid(inode,ia%ip,ia%spec,extent, &
                ia%xyz,ia%nsup,phi_i,r_int,xyz_zero)    
--- a/src/exx_kernel_default.f90
+++ b/src/exx_kernel_default.f90
@@ -1169,12 +1169,12 @@ contains
                    Ome_kj(1:2*extent+1, 1:2*extent+1, 1:2*extent+1) => Ome_kj_1d_buffer
                    !
                    call start_timer(tmr_std_exx_accumul)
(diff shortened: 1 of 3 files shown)


src/exx_kernel_default.f90
```diff
@@ -1156,14 +1160,17 @@ subroutine m_kern_exx_cri(k_off, kpart, ib_nd_acc, ibaddr, nbnab, &
...
-                   !$omp parallel do schedule(runtime) collapse(2) default(none) reduction(+: c) &
+                   !$omp parallel default(none) reduction(+: c) &
                    !$omp     shared(kg,jb,tmr_std_exx_poisson,tmr_std_exx_accumul,Phy_k,phi_j,phi_k,ncbeg,ia,tmr_std_exx_matmult,ewald_pot,phi_i, &
                    !$omp            exx_psolver,exx_pscheme,extent,dv,ewald_rho,inode,pulay_radius,p_omega,p_gauss,w_gauss,reckernel_3d,r_int) &
-                   !$omp     private(nsf1,nsf2,work_out_3d,work_in_3d,ewald_charge,Ome_kj, &
-                   !$omp             ncaddr,nsf3,exx_mat_elem,r,s,t)
+                   !$omp     private(nsf1,nsf2,work_out_3d,work_in_3d,ewald_charge,Ome_kj_1d_buffer, &
+                   !$omp             ncaddr,nsf3,exx_mat_elem,r,s,t) &
+                   !$omp     firstprivate(Ome_kj)
+                   Ome_kj(1:2*extent+1, 1:2*extent+1, 1:2*extent+1) => Ome_kj_1d_buffer
```
tkoskela, line 1172: Why is `Ome_kj` firstprivate here? 

When we enter the parallel region, `Ome_kj_1d_buffer` is made private and `Ome_kj` is an unassociated pointer. Does something break if `Ome_kj` is private? I can see that you want to point `Ome_kj` to the buffer after the private copies have been created, but before you start the parallel loop and I think that is the right way to do it.

src/exx_kernel_default.f90
```diff
@@ -1191,31 +1198,18 @@ subroutine m_kern_exx_cri(k_off, kpart, ib_nd_acc, ibaddr, nbnab, &
...
-                                     exx_mat_elem = exx_mat_elem &                                    
-                                          + phi_i(t,s,r,nsf3)    &
-                                          * Ome_kj(t,s,r)
-
-                                  end do
-                               end do
-                            end do
-                            !
-                            c(ncaddr + nsf3 - 1) = c(ncaddr + nsf3 - 1) + exx_mat_elem * dv
+                            c(ncaddr + nsf3 - 1) = c(ncaddr + nsf3 - 1) + dot((2*extent+1)**3, phi_i(:,:,:,nsf3), 1, Ome_kj, 1) * dv
```
tkoskela, line 1201:
```suggestion
                            c(ncaddr + nsf3 - 1) = c(ncaddr + nsf3 - 1) &
                               + dot((2*extent+1)**3, phi_i(:,:,:,nsf3), 1, Ome_kj, 1) * dv

```

tkoskela (commented):
The changes in this PR look pretty straightforward to me. It would be interesting to time the loop ` do nsf3 = 1, ia%nsup` with both versions of the code and see how much faster the blas call is. I guess you could do this with one thread so you don't have to worry about the thread-safety of the timers. With the graphs you've shown in the other PRs, I'm not entirely convinced this is making the code significantly faster.

connoraird:
> The changes in this PR look pretty straightforward to me. It would be interesting to time the loop ` do nsf3 = 1, ia%nsup` with both versions of the code and see how much faster the blas call is. I guess you could do this with one thread so you don't have to worry about the thread-safety of the timers. With the graphs you've shown in the other PRs, I'm not entirely convinced this is making the code significantly faster.

Results of timing nsf3 loop
| threads | 276-combine-nsf-loops | 276-use-blas
| -------- | -------------------------- | --------------
| 1	       | 2.47956 s                        | 1.95329 s
| 2	       | 1.45973 s                        | 1.24583 s
| 4	       | 0.77407 s                        | 0.67395 s
| 8	       | 0.46875 s                        | 0.39146 s
| 16	       | 0.27807 s                        | 0.22739 s

src/exx_kernel_default.f90
```diff
@@ -1156,14 +1160,17 @@ subroutine m_kern_exx_cri(k_off, kpart, ib_nd_acc, ibaddr, nbnab, &
...
-                   !$omp     shared(kg,jb,tmr_std_exx_poisson,tmr_std_exx_accumul,Phy_k,phi_j,phi_k,ncbeg,ia,tmr_std_exx_matmult,ewald_pot,phi_i, &
-                   !$omp            exx_psolver,exx_pscheme,extent,dv,ewald_rho,inode,pulay_radius,p_omega,p_gauss,w_gauss,reckernel_3d,r_int) &
-                   !$omp     private(nsf1,nsf2,work_out_3d,work_in_3d,ewald_charge,Ome_kj, &
-                   !$omp             ncaddr,nsf3,exx_mat_elem,r,s,t)
+                   !$omp parallel default(none) reduction(+: c) &
+                   !$omp     shared(kg,jb,tmr_std_exx_poisson,tmr_std_exx_accumul,Phy_k,phi_j,phi_k,ncbeg,ia, &
+                   !$omp            tmr_std_exx_matmult,ewald_pot,phi_i,exx_psolver,exx_pscheme,extent,dv,    &
+                   !$omp            ewald_rho,inode,pulay_radius,p_omega,p_gauss,w_gauss,reckernel_3d,r_int)  &
+                   !$omp     private(nsf1,nsf2,work_out_3d,work_in_3d,ewald_charge,Ome_kj_1d_buffer,Ome_kj,   &
+                   !$omp             ncaddr,nsf3,exx_mat_elem,r,s,t) &
```
tkoskela, line 1171:
```suggestion
                   !$omp             ncaddr,nsf3,exx_mat_elem,r,s,t)

```

tkoskela (approved):

PR #324, opened 2024-02-09
276 thread exx phi on grid

connoraird:
## Description
- The xyz nested loop as been thread parallised in `exx_phi_on_grid`

## Speedup plot
These plots shows the performance of test `test_004_isol_C2H4_4proc_PBE0CRI` for 1 mpi process
276-thread-phi-on-grid

### DIFF
--- a/src/exx_evalpao.f90
+++ b/src/exx_evalpao.f90
@@ -182,10 +182,7 @@ contains
                    zeta_loop: do acz = 1, pao(spec)%angmom(l1)%n_zeta_in_angmom
 
                       magn_loop: do m1 = -l1, l1                      
-
-                         pao_val = zero
-                         !y_val   = zero             
-
+                      
                          call evaluate_pao(i_dummy,spec,l1,acz,m1,x,y,z,pao_val,exx_cartesian)
 
                          ! Put pao_val directly into phi_on_grid
--- a/src/exx_evalpao.f90
+++ b/src/exx_evalpao.f90
@@ -200,6 +200,7 @@ contains
              end do grid_z_loop
           end do grid_y_loop
        end do grid_x_loop
+       !$omp end parallel do
 
     end if
 
--- a/src/exx_evalpao.f90
+++ b/src/exx_evalpao.f90
@@ -157,8 +157,8 @@ contains
        !print*,
        xyz_offset = xyz + rst
        !$omp parallel do collapse(3) schedule(runtime) default(none) & 
-       !$omp    shared(mx,my,mz,px,py,pz,grid_spacing,xyz_offset,pao,spec,phi_on_grid,i_dummy,exx_cartesian) &
-       !$omp    private(nx,ny,nz,x,y,z,count,l1,acz,m1,pao_val)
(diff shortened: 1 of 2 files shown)


src/exx_evalpao.f90
```diff
@@ -155,19 +155,21 @@ subroutine exx_phi_on_grid(inode,atom,spec,extent,xyz,nsuppfuncs,phi_on_grid,r_i
           pz = pz -ijk(3)+1
        end if overlap_box
        !print*,
+       xyz_offset = xyz + rst
+       !$omp parallel do collapse(3) schedule(runtime) default(none) & 
+       !$omp    shared(mx,my,mz,px,py,pz,grid_spacing,xyz_offset,pao,spec,phi_on_grid,i_dummy,exx_cartesian,extent) &
```
tkoskela, line 160: Shorter lines please :pleading_face:

src/exx_evalpao.f90
```diff
@@ -155,19 +155,21 @@ subroutine exx_phi_on_grid(inode,atom,spec,extent,xyz,nsuppfuncs,phi_on_grid,r_i
...
-          x = xyz(1) + real(nx,double)*grid_spacing + rst(1)
-
           grid_y_loop: do ny = my, py
-             y = xyz(2) + real(ny,double)*grid_spacing + rst(2)
-
              grid_z_loop: do nz = mz, pz
-                z = xyz(3) + real(nz,double)*grid_spacing + rst(3)
+                x = nx*grid_spacing + xyz_offset(1)
+                y = ny*grid_spacing + xyz_offset(2)
+                z = nz*grid_spacing + xyz_offset(3)
```
tkoskela, line 167: This is fine. If there's spare time, I'd like to try the following:

precompute `x`, `y`, and `z` into arrays outside of the loop. These are just 1d arrays so the memory footprint should be small, and we would avoid redundant recomputations of y and z. Could use a Array of Structures data format so that structures of x,y,z are aligned in memory (ie, `real, dimension(3,N) :: xyz`, where N = max(px-mx, py-my, pz-mz), or something like that. Then inside this loop you could just use `xyz(1,nx)`, `xyz(2,ny)`, `xyz(3,nz)`. I'm not sure if this makes much performance difference. As we learned this week, "memory is expensive, flops are free".

tkoskela (approved):
This is a nice, minimal change that gets a parallel performance boost! I like it.

src/exx_evalpao.f90
connoraird: I'm not sure this would be worth it. I've just tested the concept with a short program and there seems to be no difference. When accessing xyz in the nested loop, the different values of nx, ny and nz make the memory accessed non-contiguous so we'll be getting lots of cache misses.

tkoskela: It should be possible to arrange the data such that the nx,ny,nz accesses are contiguous. Maybe I got it wrong in my comment. In principle I agree, if it seems like this isn't worth it, let's not spend much time on it.

PR #327, opened 2024-02-13
Optimize ompGemm_m multiply kernel

tkoskela:
This PR optimizes the memory copies in `m_kern_min` and `m_kern_max` of `multiply_kernel_ompGemm_m`. The intention is that this multiply kernel should be favoured in most (if not all) cases and the worse performing multiply kernels could be dropped.

In `m_kern_max` we avoid making temporary copies of matrices `A` and `B` completely by calling `dgemm` on the whole matrices including zero elements. This is the main performance gain. The zero elements are skipped when copying the temporary result back to `C`, keeping the end result sparse. `dgemm` also accepts 1D arrays as arguments, so the only remaining work before the `dgemm` call is computing the start and end indices into `A` and `B`. After the `dgemm` call the result is copied into `C` from a temporary array.

In `m_kern_min` we still make temporary copies of `B` and `C` because both of them are stored in a sparse data structure. I've tried vectorizing the copies but the benefits are not great due to the typically small values of `nd1` and `nd3`. No temporary copy is needed for the result `A`.

The common index calculations of `m_kern_max` and `m_kern_min` have been refactored into a separate subroutine `precompute_indices` to reduce code duplication.

A number of unused variables have also been removed and comments added.

Also adds `system.myriad.make` since I've found myriad useful for testing and development


tkoskela:
Performance profiling data was collected using `benchmarks/matrix_multiply/si_442.xtl`. I ran it on myriad with 8 MPI ranks and 4 OpenMP threads using Intel(R) VTune(TM) Profiler 2022.2.0.

The main thing to note is the time spent in `m_kern_min` and `m_kern_max`, shown in the "Top Hotspots" table. As an interesting byproduct the time in `kmp_fork_barrier`, ie. the OpenMP load imbalance has significantly decreased as well. This would make sense if there are not enough `m_kern_min` and `m_kern_max` calls to keep all threads busy. The serial calls are now faster and the idle threads wait for less time. I doubt anything has fundamentally improved with the load balance.

## Baseline: Develop branch

image

## This PR

image

tkoskela:
On the topic of vectorized memory copies of `B` and `C` in `m_kern_min`, this is what the data in Advisor looks like:

image

You may note that Advisor sees three loops on lines 411 and 412. I think one is the vectorized loop, one is the remaineder (the iterations not divisible by 4), I'm not entirely sure what the third is. The important bit to see here is the `Gain` column, which shows the gain from vectorization is less than 1. In the `Trip Counts` column we see the average trip counts in the vectorized loops are 1 or 2, which explains the low gain. This data is collected using the LargerND input data set I got from @davidbowler

davidbowler (changes requested):
We need data on one thread for comparison, and running between nodes as well on single nodes.  Also need to find out what/where kmp_fork_barrier is and whether we can reduce it.

tkoskela:
Here is data on one thread. These screenshots are a bit busy but the main points here are:
- The total elapsed time is reduced from 505s to 401s
- The time spent in `dgemm` is unchanged.
- The time spent in `m_kern_max` is reduced from 105s to 42s
- The time spent in `m_kern_min` is reduced from 75s to 69s

# Develop branch

image

# This PR

image

tkoskela:
Following the advice in [this stackoverflow thread](https://stackoverflow.com/questions/43678381/what-is-kmp-fork-barrier-and-how-to-see-if-there-is-load-imbalance)

> You need to learn to use VTune better. It has specific OpenMP analyses which avoid you having to ask about the internals of the OpenMP runtime.

I ran the OpenMP threading analysis. In the screenshot below I've zoomed the graph at the bottom in to a typical 2s interval in the middle of the run, where primarily `multiply_module`, `calc_matrix_elements_module` and `pao_grid_transform_module` are being called. I probably won't be able to explain things here, but let's discuss when we next meet.

My main observation here is that the main source of wasted time is `Serial - outside parallel regions`, rather than thread imabalance in the parallel region as I previously thought. It's not clear to me why in the graph at the bottom the worker threads are sometimes red and sometimes green outside of parallel regions. Maybe this is an internal optimization in the runtime and it has some logic whether to keep the threads spinning or make them wait.

image

tkoskela:
Expaning the Serial line, you get a breakdown of where it is coming from

image

src/multiply_kernel_ompGemm_m.f90
```diff
@@ -390,8 +345,8 @@ subroutine m_kern_min(k_off, kpart, ib_nd_acc, ibaddr, nbnab,       &
     integer :: kpart, k_off
     ! Remember that a is a local transpose
     real(double) :: a(lena)
-    real(double) :: b(lenb)
-    real(double) :: c(lenc)
+    real(double), target :: b(lenb)
```
davidbowler, line 348: Why are these targets?

davidbowler (commented):
I think that this is all fine, though the performance gain I'm seeing over ompGemm_m in v1.3 is often quite small.  A few little queries (redundant targets for instance)

davidbowler:
Just for my information, have you got numbers comparing ompGemm_m in v1.3 and in this version? It's not clear from above which kernel was used for the baseline.

src/multiply_kernel_ompGemm_m.f90
tkoskela: These are left-overs from the previous version where I was using pointers. Thanks for spotting, I'll remove them.

davidbowler:
This seems like a very sensible way forward.  My only remaining question is whether we really need the explicit loop over `nd1` or whether that should be a range.

PR #331, opened 2024-02-20
Update kathleen system make file

connoraird:


tkoskela (approved):

tkoskela:
Actually, there is a problem here. XC_COMPFLAGS are defined after COMPFLAGS. Let me fix that

ilectra (approved):

PR #335, opened 2024-02-27
Changed default system.make behaviour, tidied up files

davidbowler:
As well as defaulting to system.make (no automatic hostname) I removed unneeded variables from system.make and tweaked the documentation

I tweaked the Young and Kathleen files: please check these if you have time @ilectra @tkoskela @connoraird


tkoskela (changes requested):
This all seems to work for me. Thanks for changing the documentation and the prompt in the Makefile as well! Since we are not machine-detecting the system name anymore, I would like to change the filenames to a more human-friendly format, ie.

`system.kathleen.ac.ucl.make` -> `system.kathleen.make`
`system.ib.young.ucl.ac.uk.make` -> `system.young.make`

davidbowler:
Absolutely - that's done.

tkoskela (approved):

PR #345, opened 2024-04-15
Fix error in l-m resolved pDOS

davidbowler:
Indexing for m and zeta is wrong way around when resolving pDOS onto l and m; this is now fixed.


tsuyoshi38 (approved):
Sorry, I have not checked whether this code runs correctly or not. But, the code looks correct to me.

PR #348, opened 2024-04-24
Update versions of github actions

tkoskela:
This updates the `checkout` and `setup-python` actions used in the CI workflow to their latest released versions. Silences the warnings in the github actions output.


tkoskela:
There's an [issue on GitHub actions](https://github.com/actions/runner-images/issues/9736) that is causing `apt update` to fail at the moment. This is making all of our tests fail. I hope it should get resolved quickly.

davidbowler (approved):

PR #349, opened 2024-04-25
Extract inner cri calc into separate subroutine

connoraird:
I believe the code within the `do nsf2 = 1, jb%nsup` loop of `m_kern_exx_cri` and `m_kern_exx_eri` is almost identical. Therefore, we should be able to extract this into a subroutine and therefore get the gain of using BLAS and any other optimisation in one place.

The current issue is that in `m_ker_exx_eri` there is an option to `backup_eris`. If true, this contiguously accesses a 1D array via `eris(kpart)%store_eris(count)`. With the new OpenMP thread implementation we no longer access these elements in this order.

### DIFF
--- a/src/exx_kernel_default.f90
+++ b/src/exx_kernel_default.f90
@@ -882,8 +882,10 @@ contains
        !
        !call io_close(unit_eri_debug)
        !call io_close(unit_exx_debug)
-       call io_close(unit_memory_write)
-       call io_close(unit_timers_write)
+       if (fdf_boolean('IO.WriteOutToFile',.true.)) then
+         call io_close(unit_memory_write)
+         call io_close(unit_timers_write)
+       end if
        !
        !
     end if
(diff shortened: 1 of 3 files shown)


src/exx_kernel_default.f90
```diff
@@ -1432,17 +1476,39 @@ subroutine m_kern_exx_eri(k_off, kpart, ib_nd_acc, ibaddr, nbnab, &
...
                    K_val = real(1,double)
                 else
                    K_val = b(nbaddr+nsf_kg-1)
                 end if
+                !
+                ! The current state of count
+                ! kg_count = (k - 1)      * nbnab(k_in_part) * ld%nsup * kg%nsup * at%n_hnab(k_in_halo) + &
+                !         (l - 1)                         * ld%nsup * kg%nsup * at%n_hnab(k_in_halo) + &
+                !         (nsf_ld - 1)                              * kg%nsup * at%n_hnab(k_in_halo) + &
+                !         (nsf_kg - 1)                                        * at%n_hnab(k_in_halo)
```
connoraird, line 1508:
```suggestion
                ! kg_count = (k - 1)      * nbnab(k_in_part) * ld%nsup * kg%nsup * at%n_hnab(k_in_halo) + &
                !            (l - 1)                         * ld%nsup * kg%nsup * at%n_hnab(k_in_halo) + &
                !            (nsf_ld - 1)                              * kg%nsup * at%n_hnab(k_in_halo) + &
                !            (nsf_kg - 1)                                        * at%n_hnab(k_in_halo)

```

PR #360, opened 2024-07-19
Document OpenMP scheduling

tkoskela:
I added a brief description of the OpenMP runtime scheduling into the documentation, based on our discussion today.


davidbowler:
This is fine: presumably there isn't general advice we can give on setting the value as it will be system/problem dependent...

davidbowler (approved):

PR #366, opened 2024-09-04
Documentation for installation on Ubuntu

AugustinLu:
Instructions have been added for compilation using Ubuntu.
(In #365 the latest commit had not yet been pushed to the repository..)

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -119,6 +119,7 @@ Installing on Ubuntu
 
 CONQUEST can be compiled on Ubuntu after installing the required packages. The below instructions are given for Ubuntu 22.04 LTS and Ubuntu 24.04 LTS.
 The source files will be downloaded into the ``${USER}/local/src`` directory. The ${USER} variable will be automatically replaced by the current username.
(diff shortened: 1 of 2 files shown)


davidbowler (approved):
I have checked this (using Ubuntu on a Mac) and it all works very well.  The only slight issue at present is that your suggested system.make does not set the OMP_DUMMY flag explicitly - it might be worth adding this.

davidbowler (commented):
Actually, now that I think about it, it might be better just to have a system.ubuntu.make file as part of this branch.

AugustinLu:
> I have checked this (using Ubuntu on a Mac) and it all works very well. The only slight issue at present is that your suggested system.make does not set the OMP_DUMMY flag explicitly - it might be worth adding this.

> Actually, now that I think about it, it might be better just to have a system.ubuntu.make file as part of this branch.

Thank you for your suggestions. I will add the OMP_DUMMY flag and the system.ubuntu.make file.

davidbowler:
Excellent! This looks very good - thank you.

PR #370, opened 2024-09-18
Use one system.make file for all compilations

davidbowler:
Changed PostProcessing and BasisGeneration to use the system.make file from src/system so only one is needed


tkoskela (approved):
Looks like this is replicating what we did in the main Makefile? LGTM

PR #373, opened 2024-10-18
Changed the default setting of Diag.BlockSizeR and C

tsuyoshi38:
With this version,
Default setting: Padding H = True
  Padding H = true and Diag.BlockSizeR = Diag.BlockSizeC = 32
  If users want to change the block size to check the efficiency, they should add one line
    Diag.BlockSizeR INT
when PaddingH = False
  the size of Hamiltonian is the # of matrix_size
  Diag.BlockSieR and BlockSizeC are set as in the old version of Conquest.

(Even with the present version, the default value of Padding H = 'true', though Diag.BlockSizeR or C is not defined if they are not specified.)

### DIFF
--- a/src/initial_read_module.f90
+++ b/src/initial_read_module.f90
@@ -3102,14 +3102,18 @@ contains
     ! padH or not  :temporary?   
     flag_padH = fdf_boolean('Diag.PaddingHmatrix',.true.)
 
-    if(fdf_defined('Diag.BlockSizeR')) then
+    if(flag_padH) then
+       ! default block size is 32 (we may change this value in the future)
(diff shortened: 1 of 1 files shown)


davidbowler (approved):
These changes are so small that they are numerical noise, I think.  The differences in O2 relaxation are not significant (I think that we might get something similar changing numbers of processes).

tsuyoshi38:
Dave reported a bug when Diag.PaddingHmatrix is true for small systems with multiple processes. (Thank you! > Dave)
I will issue a pull request. (but I don't know at present how I can do it..)

PR #386, opened 2024-12-17
changed the part "Diag.BlockSizeR" in manual

tsuyoshi38:
Updated a part of the manual related to  "Diag.BlockSizeR" to explain the change of the default value of BlockSizeR.


tsuyoshi38:
I am very sorry, but I have just noticed that  the default value is 32, not 30...      (Sumimasen...)

PR #393, opened 2025-01-14
Enable libxc7.0.0

tsuyoshi38:
Since libxc7.0.0 has deleted f90 interfaces, we now need to use f03 interfaces.
Instead of adding LibXC_v7_module.f90, we have decided to use f03 interfaces even in XC_LibXC_v5_module.f90 to use libXC v5 , v6 and v7.  
Users also need to change system.make to link with libXC by -lxcf03 instead of -lxcf90.
(See system.example.make)


davidbowler (approved):
Note that the BasisGeneration code uses the system.make file from the src directory so there is no update needed there.

PR #399, opened 2025-02-06
Implementing ELPA

tsuyoshi38:


tsuyoshi38:
For now, please ignore the warning messages like


> WARNING: Diag.ProcRows is not a factor of Diag.ProcCols !       6       8
> 
>  To use Cannons algorithm, np_cols must be a multiple of np_rows.
>  Switching to elpa Hermitian and scalapack
>  To use Cannons algorithm, np_cols must be a multiple of np_rows.
>  Switching to elpa Hermitian and scalapack

The first line comes from CONQUEST, while the rest from ELPA.

davidbowler (approved):

PR #405, opened 2025-02-20
Implement surface dipole correction

davidbowler:
Implement the surface dipole correction for slab calculations of Neugebauer and Scheffler (PRB *46* 16067 (1992)) and Bengtsson (PRB 59 12,301 (1999)).  Requires user to specify direction of surface normal and optionally location of dipole correction (in vacuum).  Enables output of average electrostatic potential in specified direction (with or without dipole correction).  Also added documentation and test for calculation with dipole.


tsuyoshi38 (approved):
Let me ask you one question. 
The subroutines like get_band_density in density_module.f90 are also removed, since they are now done in PostProcess, I guess. Am I correct ?

The code for surface dipole correction looks good.
(Can we also treat "charged state" ? )

PR #409, opened 2025-03-21
Implement correction to S-Pulay force and stress

davidbowler:
We were missing a contribution to the force and stress that comes from the Lagrange multiplier used to maintain electron number.  This has now been implemented.


davidbowler:
The failed test is a direct result of the bug fix which has changed the stress!  I need to post a replacement reference output.

davidbowler:
Fixed by adding an updated reference file (with correct stress)

tsuyoshi38 (approved):
With the correction, the constant of motion in NVE simulations with O(N) behaves much better. Thank you very much!

PR #434, opened 2026-02-11
Add citations for LibXC

davidbowler:
Citation for LibXC added to output file, and URL and citation added to manual (under Installation).  Added URL for ELPA at the same time.


lionelalexandre (approved):
thanks Dave for this

PR #467, opened 2026-08-07
Initialize eri_gto to zero.

### LINKED ISSUE #466: Missing initial value in EXX GTO
On mac os, the behaviour of test 005 PBE0GTO is unstable. We either get very large values (like 50000 Ha) or the calculation does not converge. The behavior is erratic and two consecutive runs yield different results.

This seems to originate from `src/exx_erigto.f90`, where we declare variable `eri_gto` but do not initialize it to zero. Therefore, it is unclear from what values it will accumulate from. Probably, the way of handling undeclared variable depends on the system (and compiler?), therefore I never saw the issue in Linux.

**Solution**: declare `eri_gto=zero` at the top of the routine.

AugustinLu:
This pull request adds zero initialization for variable `eri_gto`. 

This solves #466 

Results on Mac OS are now consistent with the reference value:
`Conquest_out:      |* Harris-Foulkes energy   =       -13.983024122321860 Ha`
`Conquest_out.ref:      |* Harris-Foulkes energy   =       -13.983024122331756 Ha`

### DIFF
--- a/src/exx_erigto.f90
+++ b/src/exx_erigto.f90
@@ -666,6 +666,8 @@ contains
     
     real(double), dimension(5) :: F
 
+    eri_gto = zero
+
     i_nt = trim(gto( i_nsp )%sf( i )%nt)
     j_nt = trim(gto( j_nsp )%sf( j )%nt)
     k_nt = trim(gto( k_nsp )%sf( k )%nt)


davidbowler:
I never saw this issue on MacOS but it's a sensible thing to do.

davidbowler (approved):

AugustinLu:
```====================================================
(base) augustin@[MyMachine] % bash compare.sh
====================================================
 Conquest Version Comparison Workflow
 Old: ./Conquest_develop in Old/
 New: ./Conquest_fixed in New/
 MPI Processes: 2
====================================================

[1/3] Running Old version...
[2/3] Running New version...
[3/3] Comparing results...
----------------------------------------------------
 Metric         | Old Version | New Version
----------------------------------------------------
 SCF Iterations | 22          | 14
 Total Energy   | **-55403.462136901252961** | **-13.983024122316245**
----------------------------------------------------

CONCLUSION: The outputs differ.
====================================================
```


Fix confirmed.

PR #481, opened 2026-09-15
Fix documentation issues and improve input/output descriptions

AugustinLu:
I have audited the documentation files and proposed a several changes to remove obsolete information, correct a few keywords and ensure that the information is up to date with release 1.6.

The commits are grouped in four categories (in chronological order):
1) Typo and minor errors
2) Obsolete keywords to remove
3) Existing information that needs to be updated
4) New information specific to version 1.6 (mainly LDA+U and DOS-related information)

The source code itself was left untouched. Only README.md and files in docs folder were modified.


docs/strucrelax.rst
```diff
@@ -82,12 +82,12 @@ For large initial forces or problematic cases where the relaxation algorithms fa
 downhill search direction, it may be worth trying quenched molecular dynamics,
 which propagates the equations of motion following a simple NVE
 approach, but resets the velocities to zero when the dot product of
-force and velocity is zero.
+force and velocity becomes negative.
```
AugustinLu, line 85: Coming from 

```! Accumulate system-wide F.v and set v=0 if F.v<0
...
if(vf<zero) then
   ...zeroing message...
   v(:,:) = zero
else
   ...continue with velocity update...
end if```

File: docs/errors.rst [hunk 2]
```diff
@@ -3,8 +3,3 @@
 ===========
 Error codes
 ===========
-
```
AugustinLu, line 6: We may keep the text, but I feel it is better to remove it and start populating the section.

chpxu:
The external tools page (references to specific python scripts in `tools` directory): https://conquest.readthedocs.io/en/develop/ext-tools.html seems slightly out of date. From what I can find some of these scripts seem to be in `src/utilities`?

There is also a typo near the end
"ASE is a versatil tool to manage CONQUEST calculations without pain either:", should be versatile

docs/errors.rst
davidbowler: I agree that we should remove the text.

AugustinLu:
Thank you for your comments.

> The external tools page (references to specific python scripts in `tools` directory): https://conquest.readthedocs.io/en/develop/ext-tools.html seems slightly out of date. From what I can find some of these scripts seem to be in `src/utilities`?

Let me check that. So far, I have been focusing on the main source code (Conquest, MakeIonFiles and PostProcessCQ). If changes are not too big, we can wrap them in the present branch as well.

> There is also a typo near the end "ASE is a versatil tool to manage CONQUEST calculations without pain either:", should be versatile

This typo should have been corrected in commit e0c89fd. Could you confirm it?

docs/errors.rst
AugustinLu: Alright👍.
Then we can keep the committed version, which does that removal.

AugustinLu:
A bit concerning: The md_analysis script crashes due to an indentation error:

```python md_analysis.py
  File "[...]/src/utilities/md_analysis.py", line 281
    ax1.plot(data['time'][opts.nskip:], data['H\''][opts.nskip:],
IndentationError: unexpected indent
```

It must be a trivial error. But I would not fix it here because the present branch is dedicated to the documentation.

EDIT: After checks, it is not trivial. Some scipy calls are outdated and some other parts need corrections. In this branch, I chose to not alter the main codebase, so it will be dealt with in another issue/pull request. The documentation edits proposed here will assume that the script works and does as announced.

docs/theory-md.rst
```diff
@@ -281,17 +281,17 @@ Weak coupling thermostat/barostat
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Instead of modifying the Hamiltonian, the Berendsen-type weak coupling method
-:cite:`t-Berendsen1984` involves coupling the ionic degrees of freedom to a an
+:cite:`t-Berendsen1984` involves coupling the ionic degrees of freedom to an
```
AugustinLu, line 284: It seems there is no more Berendsen nor Langevin dynamics in the source code as of now. Do we plan to re-introduce them in the future?

davidbowler: Berendsen was removed because it has fundamental problems (transfers energy from vibrational to translational modes).  Langevin dynamics might return though the stochastic velocity rescaling (SVR) serves a similar purpose.

davidbowler:
I think that the MD analysis scripts are rather outdated, and only one of the three scripts mentioned are actually present in the distribution (and in the directory `src/utilities` as @chpxu points out).  I wonder whether we should remove this section, or at least reduce it, so that it describes only the relevant files.

docs/theory-md.rst
AugustinLu: I see. Then we can leave the description as is, and link it to relevant keywords when they are effectively implemented.

AugustinLu:
> I think that the MD analysis scripts are rather outdated, and only one of the three scripts mentioned are actually present in the distribution (and in the directory `src/utilities` as @chpxu points out).  I wonder whether we should remove this section, or at least reduce it, so that it describes only the relevant files.

Actually, multiple analyses are in the same script, and which one is chosen depends on what we set as arguments when running the script. I will issue a fix for them in a separate branch to at least have a functional script. The latest commits assume that the scripts have been fixed. 

At a later stage (for ver. 1.7?), we may provide updated scripts.

davidbowler:
I think that this documentation is ready to be merged into release-1.6 so please convert to a full pull request.

davidbowler (approved):

tsuyoshi38 (approved):

